PR #39, opened 2020-10-26
DEVDOCS-2363: Traci707-modified-channels-v3

traci707:
Removed the callout below as it is incomplete. Possibly the other properties were added at a later time.  Replaced this information with a definition of the endpoint. The request body schema indicates the supported properties which should be good for our readers.

Currently only the following properties can be updated on channel listings:

state
variants.state

### DIFF
--- a/reference/channels.v3.yml
+++ b/reference/channels.v3.yml
@@ -563,13 +563,7 @@ paths:
       tags:
         - Channel Listings
       description: >-
-        Currently only the following properties can be updated on channel
-        listings:
-
-
-        * state
-
-        * variants.state
+        Updates a listing for provided channel
     post:
       summary: Create Channel Listings
       operationId: createChannelListings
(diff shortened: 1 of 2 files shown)


reference/channels.v3.yml
```diff
@@ -452,11 +452,16 @@ paths:
...
         - Channel Listings
+
+      description: >-
+        Updates a listing for provided channel
+
       description: |-
         Currently only the following properties can be updated on channel listings:
 
         * state
         * variants.state
```
aglensmith, line 463: Need to remove this duplicate description.

aglensmith (changes requested):

reference/channels.v3.yml
traci707: I don't see a duplication.

traci707: I just saw that I had conflicts in the file.  I took care of them.

PR #40, opened 2020-10-27
Devdocs-2365 get product by

traci707:
Removed price_list_id query parameter from Get a Product endpoint because it no longer works.
See [DEVDOCS-2365](https://jira.bigcommerce.com/browse/DEVDOCS-2365)

### DIFF
--- a/reference/channels.v3.yml
+++ b/reference/channels.v3.yml
@@ -563,13 +563,7 @@ paths:
       tags:
         - Channel Listings
       description: >-
-        Currently only the following properties can be updated on channel
-        listings:
-
-
-        * state
-
-        * variants.state
+        Updates a listing for provided channel
     post:
       summary: Create Channel Listings
       operationId: createChannelListings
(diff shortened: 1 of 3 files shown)


reference/channels.v3.yml
```diff
@@ -452,11 +452,9 @@ paths:
                 1.product_id: error.path.missing
```
aglensmith, line 452: Remove the changes to this file from the pr.

reference/marketing.v2.yml
```diff
@@ -2,61 +2,37 @@ swagger: '2.0'
 info:
```
aglensmith, line 2: remove the changes to this file from the pr.

aglensmith (changes requested):

reference/channels.v3.yml
traci707: When I remove it, I receive a message saying the yml file is invalid.

reference/marketing.v2.yml
traci707: This change also invalidates the file.

PR #53, opened 2020-11-23
DEVDOCS-2407-add-date-option-create-carts.v3.yml

traci707:
DEVDOCS-2407 indicates that documentation is missing for the case of creating a cart using the date option. I added in an additional example so users would know how to do this.

### DIFF
--- a/reference/carts.v3.yml
+++ b/reference/carts.v3.yml
@@ -81,6 +81,22 @@ paths:
                       }
                   ]
               }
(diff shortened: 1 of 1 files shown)


mjhmatt:
Looks fine to me, is it worth adding we also support sending day/month/year?
`"option_value": {    "day" : "01",    "month" : "02",    "year" : "2020"    } `

aglensmith (approved):

PR #79, opened 2020-12-21
Fix address2 and company example values in checkout V3 API

TomA-R:
Fix address2 and company example values in checkout V3 API


rabinshr (approved):

ektravel (approved):
LGTM

PR #90, opened 2021-01-06
Updated operation ids and descriptions

ektravel:


aglensmith (changes requested):
Stoplight Next uses the operation ID to generate the page's URL. Changing the Operation ID will result in the old URL to 404. We should leave the IDs unchanged for any published pages.

PR #108, opened 2021-01-15
DEVDOCS-2236

ektravel:
Updated query parameters for Channels & PriceLists APIs

### DIFF
--- a/reference/channels.v3.yml
+++ b/reference/channels.v3.yml
@@ -114,7 +114,10 @@ paths:
           $ref: '#/responses/channel_Resp_Collection'
         '500':
           $ref: '#/responses/error_Resp'
-      description: Returns a list of *Channels*. Will always return the BigCommerce storefront with an ID of `1`.
+      description: |-
+        Returns a list of *Channels*. 
+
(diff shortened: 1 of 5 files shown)


reference/channels.v3.yml
```diff
@@ -116,12 +116,42 @@ paths:
           $ref: '#/responses/error_Resp'
       description: Returns a list of *Channels*. Will always return the BigCommerce storefront with an ID of `1`.
```
amckemie, line 117: This is not something that you were supposed to do, but I'm realizing we could improve as I read this, but let's change this to "will always return the default BigCommerce storefront with an ID of `1`". The default storefront is always the first one that is created when a BigCommerce store is provisioned. We may want to add that in there, but I'll let you wordsmith that :)

amckemie (approved):

amckemie:
@ektravel Apologies, but I just realized we also missed the filter for "platform". This is very similar to "type" in that you can pass in any of the accepted platforms.

reference/channels.v3.yml
ektravel: @amckemie Submitted changes on [PR 111](https://github.com/bigcommerce/api-specs/pull/111).

PR #144, opened 2021-02-03
DEVDOCS-2517: Update Product Image Types

ektravel:
Updated supported image types


markcmurphy (approved):
I'm surprised XBM is on this list because it's a security threat that browsers haven't supported in years, but I was able to successfully upload one (though Chrome won't display it).

PR #171, opened 2021-02-16
Fix Incorrect Float Type in Orders V2

ricokahler:
I was trying to run the Orders V2 schema through [`dtsgenerator`](https://github.com/horiuchi/dtsgenerator) and got the following error:

```
Error: unknown type: float
```

After closer inspection, I think you may have switched the `number` and `float` `type` and `format` respectively.

See [here](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#dataTypes) for the Open API spec. `float` is not a type.

This fixes the error and allows `dtsgenerator` to work.

---

Hope this helps! I tried to follow PR conventions as much as I could.

### DIFF
--- a/reference/orders.v2.oas2.yml
+++ b/reference/orders.v2.oas2.yml
@@ -2732,8 +2732,8 @@ components:
                 type:
                  - string
                  - integer
-                 - float
-                format: number
+                 - number
+                format: float
               wrapping_cost_ex_tax:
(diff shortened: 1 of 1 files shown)


ricokahler:
Related: https://github.com/horiuchi/dtsgenerator/pull/404#issuecomment-609526278

aglensmith (approved):
Good catch, and thanks for the PR.

PR #200, opened 2021-03-01
feat(orders): ORDERS-3998 add coupon code validation so only white listed values are allowed

bc-vincent-zhao:
As part of security fix to prevent XSS using coupon code, we need to add white list validation on the coupon code input. (see https://jira.bigcommerce.com/browse/ORDERS-3972 for details)

Given this change may come as a potential breaking change (we might have api users that use other special characters that are outside the whitelist), I would really appreciate guidance from the dev-doc team what's the best way to give our api users notification about this change.

cc @bigcommerce/dev-docs

### DIFF
--- a/reference/marketing.v2.yml
+++ b/reference/marketing.v2.yml
@@ -934,8 +934,9 @@ definitions:
         type: boolean
       code:
         type: string
-        description: The coupon code that customers will use to receive their discounts. Value must be unique.
+        description: The coupon code that customers will use to receive their discounts. Value must be unique. Only alphanumeric values, white space, underscore and hypen are allowed.
         example: S2549JM0Y
+        pattern: '[a-zA-Z0-9_\ -]'
       applies_to:
         type: object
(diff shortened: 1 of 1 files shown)


uwikaiddi:
@bc-vincent-zhao, thank you for reaching out to Dev Docs on this matter. Currently, we use the Dev Center changelog to alert users of potentially breaking changes. We tag them as "warning" changes. 

However, I would recommend giving users some time to make preemptive changes beforehand. We would alert them the same way, by making a changelog announcement. Would giving users some time to prepare be possible?

bc-vincent-zhao:
> Would giving users some time to prepare be possible?

@uwikaiddi , yes definitely.  I'm working with PM to figure out the right timeframe and gives enterprise clients head up through their account managers.

uwikaiddi:
@traci707, here is the associated changelog ticket: https://jira.bigcommerce.com/browse/DEVDOCS-2575

PR #201, opened 2021-03-01
feat(orders): ORDERS-3998 add coupon code validation so only whitelis…

bc-vincent-zhao:
…ted values are allowed

As part of security fix to prevent XSS using coupon code, we need to add white list validation on the coupon code input. (see https://jira.bigcommerce.com/browse/ORDERS-3972 for details)

Given this change may come as a potential breaking change (we might have api users that use other special characters that are outside the whitelist), I would really appreciate guidance from the dev-doc team what's the best way to give our api users notification about this change.

Same as https://github.com/bigcommerce/api-specs/pull/200, just as draft PR to prevent accidental merge

cc @bigcommerce/dev-docs

### DIFF
--- a/reference/marketing.v2.yml
+++ b/reference/marketing.v2.yml
@@ -933,8 +933,9 @@ definitions:
         type: boolean
       code:
         type: string
-        description: The coupon code that customers will use to receive their discounts. Value must be unique.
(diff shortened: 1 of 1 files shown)


bc-vincent-zhao:
ping @ektravel

ektravel (approved):

PR #218, opened 2021-03-03
Update Checkout query params

ektravel:
Updated query params per [DEVDOCS-2581](https://jira.bigcommerce.com/browse/DEVDOCS-2581) and [DEVDOCS-2582](https://jira.bigcommerce.com/browse/DEVDOCS-2582).

### DIFF
--- a/reference/checkouts.sf.yml
+++ b/reference/checkouts.sf.yml
@@ -43,6 +43,17 @@ paths:
           in: query
           type: string
           default: consignments.availableShippingOptions
+          description: |-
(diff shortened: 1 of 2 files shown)


aglensmith (approved):
Approving, but we should also specify enum type and list allowed values like we do for include param in carts.sf.yml:

```
    include:
      name: include
      in: query
      description: Include product options in specified line item types.
      schema:
        type: string
        enum:
          - lineItems.physicalItems.options
          - lineItems.digitalItems.options
          - 'lineItems.digitalItems.options,lineItems.physicalItems.options'
```

PR #229, opened 2021-03-11
DEVDOCS-2612 remove unnecessary fields from POST / PUT order shipment request bodies

bc-rogerqian:
## What
**1.** Updated the order shipments documentation POST request example and schema body by removing the **product_id** field.

**2.** Updated the order shipments documentation PUT request example and schema body by removing the **items** object.
## Why
**1.** When **product_id** field is included in the **POST /orders/{order_id}/shipments request**, users receive the following error:
>"status": 400, "message": "The field 'product_id' cannot be written to. Please remove it from your request before trying again."

When **product_id** field is removed from the request body and all the other fields are valid, a successful response is returned.

**2.** When **items** object is included in the **PUT/orders/{order_id}/shipments/{shipment_id}**, users receive the following error:
>"status": 400, "message": "The field 'items' cannot be written to. Please remove it from your request before trying again."

When **items** object is removed from the request body and all the other fields are valid, a successful response is returned.
## Testing/Proof
**1. POST /orders/{order_id}/shipments request**
### Before
Screen Shot 2021-03-11 at 11 40 36

### After
Screen Shot 2021-03-11 at 11 37 44

**2. PUT /orders/{order_id}/shipments/{shipment_id}**
### Before
image

### After
image

ping @bigcommerce/shipping @bigcommerce/orders @bigcommerce/dev-docs @aglensmith

### DIFF
--- a/reference/orders.v2.oas2.yml
+++ b/reference/orders.v2.oas2.yml
@@ -603,7 +603,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/orderShipment_Post_Request'
+              $ref: '#/components/schemas/orderShipment_Post'
         required: true
         x-examples:
           application/json:
@@ -676,7 +676,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/orderShipment_Post_Response'
+              $ref: '#/components/schemas/orderShipment_Put'
         required: true
         x-examples:
           application/json:
@@ -3763,9 +3763,9 @@ components:
         - product_id
         - product_options
         - quantity
-    orderShipment_Post_Request:
+    orderShipment_Post:
       type: object
-      title: orderShipment_Post_Request
+      title: orderShipment_Post
       properties:
         order_address_id:
           description: ID of the desired `shipping_address` associated with the shipment.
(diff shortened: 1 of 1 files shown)


WillemHoman (approved):
👏

agchan12 (approved):

reference/orders.v2.oas2.yml
```diff
@@ -3763,9 +3763,61 @@ components:
...
+          items:
+            type: object
+            properties:
+              order_product_id:
+                type: integer
+                example: 5
+              quantity:
+                type: number
+                example: 2
+    orderShipment_Post_Response:
```
dornah, line 3818: Does it need to be called orderShipment_Put_Request instead?

bc-rogerqian: I don't think so, because this component relates to the response body that is referenced on line 679:

`$ref: '#/components/schemas/orderShipment_Post_Response'`.

Currently we only have one component **orderShipment_Post_Response**, so I split it into two components to address the **product_id** field not being required in the request body but should continue to be returned in the response body.

dornah: Isn't line 679 the PUT request body? 🤔 
Screen Shot 2021-03-11 at 3 47 42 pm

bc-rogerqian: You're right, good pick up I didn't see that, thank you!

Changed the approach after also exploring the PUT. Noticed that the expected request body for POST and PUT are different so updated the **orderShipment_Post** component by removing the **product_id** field.

And introduced an **orderShipment_Put** component that does not contain the **items** object.

agchan12 (approved):

reference/orders.v2.oas2.yml
dornah: Awesome. Thanks Roger 👍

dornah (approved):
Thanks for fixing it  👏

PR #252, opened 2021-03-26
DEVDOCS-2656-required-queries

traci707:
[DEVDOCS-2656](https://jira.bigcommerce.com/browse/DEVDOCS-2656)
Removed 
Required Query
- consignments.available_shipping_options

Under Query Parameter changed 
default: consignments.available_shipping_options
to
default: available_shipping_options

### DIFF
--- a/reference/checkouts.v3.yml
+++ b/reference/checkouts.v3.yml
@@ -196,7 +196,7 @@ paths:
         IsMultiContentStreaming: false
   '/checkouts/{checkoutId}/consignments':
     post:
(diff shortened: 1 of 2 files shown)


reference/checkouts.v3.yml
```diff
@@ -265,13 +265,12 @@ paths:
...
         - name: include
           in: query
-          required: false
           x-enum-elements:
             - name: Enum_consignments.availableShippingOptions
               description: ''
           type: string
           description: Must be included to get available shipping options
-          default: consignments.available_shipping_options
+          default: available_shipping_options
```
snaderiBC, line 273: I think `consignments.available_shipping_options` is correct and doesn't need to change.
Basically, if `?include=consignments.availableShippingOptions` is added to the request query param the response will contain list of available shipping options. Otherwise, response will not have it.

traci707: Thank you Sepehr, I am glad you saw this. I will communicate you comments to the requestor.

bc-hnbarr:
But it is not defaulted, right? @snaderiBC 
I believe this is the confusion as it is assumed to be applied by default. Which it is not, if I understood you correctly when you sent this last week: 
`$this->defaultIncludes = [
    'selected_shipping_option',
    'shipping_address',
    'coupon_discounts',
    'discounts',
];
$this->availableIncludes = [
    'available_shipping_options',
];`
I see that you've updated this to show 'available_shipping_options' rather 'consignments.available_shipping_options,' I am thinking that the term 'default' needs to be removed, as well. What do you think?

@traci707 I believe we also need to remove the 'required query' section in the description at the top of the page.

snaderiBC:
@bc-hnbarr 
>But it is not defaulted, right?

That is correct. If it is not included explicitly, the value will not be present in the response.

>I believe this is the confusion as it is assumed to be applied by default. 

But I think we are following the Swagger definition of `default` here: https://swagger.io/docs/specification/describing-parameters/#default

The code I sent was to show that `available_shipping_options` is not part of response by default (meaning, if not explicitly specified in request, it will be absent from response).

PR #253, opened 2021-03-28
feat(shipping): TAX-1014 Tax Class update

bc-lorentz:
## What
* Add created_at and updated_at response body changes, as introduced by SHIPPING-1013.

## Why
* Due to new date time column additions from SHIPPING-1013, documentation for tax_classes is out of date, and this PR is the fix for it.

### DIFF
--- a/reference/tax_classes.v2.yml
+++ b/reference/tax_classes.v2.yml
@@ -67,15 +67,25 @@ paths:
             application/json:
               - id: 1
                 name: Non-Taxable Products
+                created_at: '1973-01-20T21:34:57.903Z'
+                updated_at: '1990-12-30T00:29:23.515Z'
               - id: 2
                 name: Shipping
+                created_at: '1973-01-20T21:34:57.903Z'
+                updated_at: '1990-12-30T00:29:23.515Z'
               - id: 3
                 name: Gift Wrapping
+                created_at: '1973-01-20T21:34:57.903Z'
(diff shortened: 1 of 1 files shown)


reference/tax_classes.v2.yml
```diff
@@ -142,6 +158,18 @@ definitions:
         description: The name of the tax class.
         example: Shipping
         type: string
+      created_at:
+        description: |
+          Date and time of the metafield's creation. Read-Only.
```
theromulans, line 163: metafield?

bc-lorentz: Good catch, thanks!

reference/tax_classes.v2.yml
```diff
@@ -142,6 +158,18 @@ definitions:
         description: The name of the tax class.
         example: Shipping
         type: string
+      created_at:
+        description: |
```
theromulans, line 162: As far as I can tell we only need this `|` here when there is a multi-line description, but both of these descriptions can be single line.

Or perhaps I don't have the right understanding - thoughts?

bc-lorentz: Easy enough to just make it single line indeed.

agchan12 (approved):

agchan12:
Might want to change the ticket number to TAX-1014

theromulans (approved):
Thank you sir. As discussed, I don't think there is any timing concerns here - let's get it out there.

PR #264, opened 2021-04-02
DEVDOCS-2605-updated_payment_methods_and_processing

traci707:
[DEVDOCS-2605](https://jira.bigcommerce.com/browse/DEVDOCS-2605)
Added a new file to replace Get All Payment Methods.  This file is payments_methods.v3.yml.
This endpoint is now based on the order_id. I updated the response slightly to match the response I received.
[https://developer.bigcommerce.com/api-reference/store-management/payment-methods-api/payment-methods/getpaymentsmethods](https://developer.bigcommerce.com/api-reference/store-management/payment-methods-api/payment-methods/getpaymentsmethods)

Updated payment_processing.yml. I updated the request body.


aglensmith (changes requested):
>Added a new file to replace Get All Payment Methods. This file is payments_methods.v3.yml.
This endpoint is now based on the order_id.

Remove payment_methods.v3.yml from the PR. This endpoint for getting accepted payment methods by order_id for payment processing is already documented [here](https://developer.bigcommerce.com/api-reference/store-management/payment-processing/accepted-methods/paymentsmethodsget), and doesn't need a new spec file. `payment_methods.v2.yml` serves a different purpose, so we shouldn't remove it.

PR #293, opened 2021-04-16
docs(orders): DEVDOCS-2753 update refund quote adjustment field

bc-rogq:
# [DEVDOCS-2753](https://jira.bigcommerce.com/browse/DEVDOCS-2753)

## What
Update the adjustment field description included in the refund quote endpoint to:

>"A negative or positive 2 decimal place rounded value that represents the difference between the refund amount requested in the refund quote and the actual amount that is refundable on the order. This value is negative when the refund amount requested in the refund quote is more than the total refundable amount. This value is positive when the total refundable amount has increased, e.g. as a result of rounding."

Update the title to:

>"Adjustment Amount"

## Why
The previous description was inaccurate, the title needs to be updated so it's consistent with changes in this PR.

### Before
image

### After
image

Revert this PR.

ping @uwikaiddi @bigcommerce/orders


reference/orders.v3.yml
```diff
@@ -2280,9 +2280,15 @@ definitions:
...
-    description: A non-negative 2 decimal place rounded value that represents that amount that can be charged/refunded via payment providers
+    description: A non-negative 2 decimal place rounded value that represents the amount that can be charged/refunded via payment providers.
     example: 1.99
     title: Amount
+  AdjustmentAmount:
+    type: number
+    format: float
+    description: A negative or positive 2 decimal place rounded value that represents the difference between the refund amount requested in the refund quote and the actual amount that is refundable on the order. This value is negative when the refund amount requested in the refund quote is more than the total refundable amount. This value is positive when the total refundable amount has increased, e.g. as a result of rounding.
+    example: -10.20
+    title: Amount
```
bc-vincent-zhao, line 2291: should title be `adjustment amount` ?

bc-vincent-zhao (approved):

reference/orders.v3.yml
bc-rogq: Updated in latest [commit](https://github.com/bigcommerce/api-specs/pull/293/commits/ffa781f12ede6bb409ef7ffa92faf45535a27674).

aglensmith (approved):

PR #314, opened 2021-04-29
docs(orders): Improve item_id description

bc-rogq:
## What
Removed **item_id** from the refund quote and create refund request endpoints when the request is for an order-level refund.

Improved the **item_id** description for refund quote and create refund response endpoints so that they speak to order product ID, order address ID and order ID and when each value is returned based on **item_type** provided.

## Why
Feedback from partner that it's duplicative and confusing to provider order ID as the **item_id** when performing a refund quote or when creating a refund.

The **item_id** description wasn't clear whether order product ID, order address ID or order ID is returned based on **item_type** provided.

### Before
Request:
image
Response:
image

### After
Request:
image

Response:
image

Revert this PR.

ping @uwikaiddi @bigcommerce/orders


reference/orders.v3.yml
```diff
@@ -2337,7 +2334,7 @@ definitions:
           - ORDER
       item_id:
         type: integer
-        description: Reference to the item
```
bc-vincent-zhao, line 2340: 👏  👍

reference/orders.v3.yml
```diff
@@ -2253,9 +2253,6 @@ definitions:
         description: Reason for the refund.
         minLength: 0
         maxLength: 1000
-      item_id:
```
bc-vincent-zhao, line 2256: is this a desire to change and require engineering work rather than a simple doc update ? 

I did a quick check and it seems like we do currently send `item_id` in the request for order level refund.

WillemHoman: We send it but it is only used here from https://github.com/bigcommerce/bigcommerce/blob/08d84bb67d250c7624502ed76d8e0bafb1a5cacf/app/Order/DomainModel/Refund/ItemsRefunder.php#L90

It's pulling the `order_id` from the url. 

We should probably update the linked code to use the order id 
Below is an example, the refund succeeds, the `item_id` does not exist 
Screen Shot 2021-05-07 at 11 14 06 am

reference/orders.v3.yml
```diff
@@ -2337,7 +2334,7 @@ definitions:
           - ORDER
       item_id:
         type: integer
-        description: Reference to the item
+        description: order_product.id corresponding to the item_types of PRODUCT, GIFT_WRAPPING. order_address.id corresponding to the item_types of SHIPPING, HANDLING. order.id corresponding to the item_type of ORDER.
```
WillemHoman, line 2337: This is definitely an improvement
You have more freedom here if you go with the block syntax using `|-`
see https://github.com/WillemHoman/api-specs/blob/d1ce2955924ee71e63604e8ca5e193fe8c4ea743/reference/orders.v3.yml#L73

You could then probably even insert a table for this like

https://github.com/WillemHoman/api-specs/blob/d1ce2955924ee71e63604e8ca5e193fe8c4ea743/reference/orders.v3.yml#L27

reference/orders.v3.yml
bc-vincent-zhao: @WillemHoman thanks for the explanation and raising a bug ticket for this. 

My guess is we had the field in the past for "consistency" reasons. If there's actually no need for that field at all as we have the `order_id` from url anyway, we should definitely explore whether we can simply remove it. 

All I tried to understand is whether this change can be merged and published now, or it needs to wait until some engineering work is done.

WillemHoman: I think it can be merged as-is, but we should address the `item_id` in the response payload as described above to make it the `order_id` regardless of whether `item_id` is passed in as part of the request payload. @bc-rogq ?

uwikaiddi:
@bc-rogq Checking back in on this ticket. Is this ready to move out of Draft and be merged?

uwikaiddi:
@WillemHoman @bc-vincent-zhao, is this change ready to be merged?

WillemHoman:
@uwikaiddi yes I think it can be merged.

uwikaiddi:
Thanks, @WillemHoman!

PR #317, opened 2021-04-30
DEVDOCS-2773: Updated thumbnail property

ektravel:


reference/widgets.v3.yml
```diff
@@ -1231,11 +1231,19 @@ definitions:
...
+          type:
+            type: string
+            example: image
+          valueKey:
+            type: string
+            example: imageUrl.src
       schema:
         type: array
         description: The schema used for each element in the array.
+        items: {}
```
bc-jcha, line 1246: I think this is `widgetSchemaTab|widgetSchemaArray`

ektravel: @bc-jcha This is `widgetSchemaArray`. The `items` keyword is required in arrays per [Swagger documentation](https://swagger.io/docs/specification/data-models/data-types/#array). The value of `items` is a schema that describes the type and format of array items.

bc-jcha: Per our offline discussion, api-specs repo already has the items empty, so we will proceed with the current code for now.

bc-jcha (approved):

reference/widgets.v3.yml
aglensmith: Merging, but we should specify the type for items.

PR #329, opened 2021-05-07
Added reference/cookie-consent.yml

markcmurphy:
https://jira.bigcommerce.com/browse/DEVDOCS-2587


aglensmith:
@ektravel for consistency, and since this is for scripts and not just cookies, let's change the filename to `consent.sf.yml`

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
+openapi: 3.0.1
+servers:
+  - url: 'https://{$$.env.store_domain}/api/storefront'
+info:
+  title: Storefront Cookie Consent API
```
ektravel, line 5:
```suggestion
  title: Storefront Cookie Consent
```
Remove "API" to match the rest of the API Reference documentation.

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
...
+  description: Storefront API for specifying consent to particular categories of cookies for compliance purposes.
```
ektravel, line 6: Update description to match the style of the API Reference documentation.
```suggestion
  description: Manage customer consent to particular categories of cookies for compliance purposes.
```

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
...
+  version: Storefront
+tags:
+  - name: Customers
```
ektravel, line 9: Is "Customers" the correct tag to use for this API? "Cookie Consent" seems to be more appropriate. Can you please verify?

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
...
+paths:
+  /consent:
+    post:
+      summary: Specify Shopper Cookie Consent Preferences
```
ektravel, line 13: Update operation description to match the style of the API Reference documentation.
```suggestion
      summary: Set Cookie Consent Preferences
```

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
...
+      tags:
+        - Customers
```
ektravel, line 15: See previous comment about tags.

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
...
+              $ref: '#/components/schemas/ConsentPreferences'
+            examples:
+              Example:
+                value:
+                  allow:
+                    - 2
+                    - 3
+                  deny:
+                    - 4
+        description: Data sent to the customer endpoint when creating a customer during checkout
```
ektravel, line 34:
```suggestion
        description: Data sent to the customer endpoint when creating a customer during checkout.
```
I think we should add more details here. Are we talking about the Storefront Customers API, Customers V2, or Customers V3. Either adding the endpoint itself (/v3/customers)or a link to the API would be good.

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
...
+        required: true
+      description: |-
```
ektravel, line 36: I think we need to fix the formatting a bit. Consent categories are not displaying as a list. Also, please remove passive voice where possible.

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
...
+      type: object
+      title: ConsentPreferences
+      description: |-
+        List of allowed and denied consent categories. Must be populated with a complete set of allowed and denied categories.
+
+        Configurable categories are:
+
+        2 - Functional
+        3 - Analytics
+        4 - Targeting/Advertising
```
ektravel, line 64:
```suggestion
        4 - Targeting/Advertising
```
Is "Targeting/Advertising" correct? Is that how a user would input it? Or should it be Targeting, Advertising? Can you please verify?

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
...
+
+        See https://developer.bigcommerce.com/api-docs/store-management/scripts for further definition of these categories.
```
ektravel, line 66:
```suggestion
        See https://developer.bigcommerce.com/api-docs/store-management/scripts for further definition of these categories. For further definition of these categories, see [Scripts API](https://developer.bigcommerce.com/api-docs/store-management/scripts).
```

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
...
+      properties:
+        allow:
+          type: array
+          description: 'Explicitly allowed consent categories. Allowed values are 2, 3, 4.'
```
ektravel, line 70: Can we remove the word "explicitly"?

reference/cookie_consent.yml
```diff
@@ -0,0 +1,94 @@
...
+          items:
+            type: integer
+            enum:
+              - 2
+              - 3
+              - 4
+            example: 3
+        deny:
+          type: array
+          description: 'Explicitly allowed consent categories. Allowed values are 2, 3, 4.'
```
ektravel, line 80: Can we remove the word "explicitly"?

ektravel (commented):
Added my suggestions.

reference/cookie_consent.yml
aglensmith: I think `Specify Shopper Cookie Consent Preferences` is better since `manage` suggests that the operation being performed is from the management perspective (i.e. user perspective) instead of the shopper perspective.

reference/cookie_consent.yml
aglensmith: Take should be `Consent`.

reference/cookie_consent.yml
markcmurphy: Updated to Targeting; Advertising

reference/cookie_consent.yml
markcmurphy: @ektravel The spec file describes them as explicit because all users implicitly agree to consent category 1: "Essential — These cookies are essential for the site to function and for any requested services to work, but do not perform any additional or secondary function. Shoppers cannot opt-out of these cookies."

Do you still think we should remove it?

ektravel: We can keep the word "explicitly".

reference/consent.sf.yml
```diff
@@ -0,0 +1,92 @@
...
+          items:
+            type: integer
+            enum:
+              - 2
+              - 3
+              - 4
+            example: 3
+        deny:
+          type: array
+          description: 'Explicitly allowed consent categories. Allowed values are 2, 3, 4.'
```
ektravel, line 78: Can you please verify that the property description is correct? It is identical to the `allow` property.

ektravel (commented):
Added one suggestion.

reference/consent.sf.yml
markcmurphy: @ektravel Good catch, just updated

ektravel (approved):

PR #331, opened 2021-05-10
DEVDOCS-2790-attribute-value-customers.v3.yml

traci707:
[DEVDOCS-2790](https://jira.bigcommerce.com/browse/DEVDOCS-2790)

Removed inconsistencies in the [Upsert Customer Attribute](https://developer.bigcommerce.com/api-reference/store-management/customers-v3/customer-attribute-values/customersattributevaluesput) endpoint by changing `value` to `attribute_value` in the Request body.  This parameter now matches the response in this endpoint and also in the [Get All Customer Attributes](https://developer.bigcommerce.com/api-reference/store-management/customers-v3/customer-attribute-values/customersattributevaluesget) endpoint.  See screenshot in the ticket.


aglensmith:
@traci707 Based on your screenshot, `value` is used for the upsert request and `attribute_value` is returned in the response. Is this correct? If so, then based on [the preview of your branch here in stoplight platform](https://bigcommerce.stoplight.io/docs/api-reference/branches/DEVDOCS-2790-attribute-value-customers.v3.yml/reference/customers.v3.yml/paths/~1customers~1attribute-values/put), this PR actually switches that around, which I don't think is correct.

aglensmith (approved):

PR #334, opened 2021-05-12
DEVDOCS-2648-create-storefront_tokens.v3.yml

traci707:
[DEVDOCS-2648](https://jira.bigcommerce.com/browse/DEVDOCS-2648)

The requestor indicated he was confused with the example test request as it contained environment variables. I replaced the environment variables with real examples similar to other parts of our docs to make it clear.

In addition, I resolved the two errors in the file. Although `allowed_cors_origins` is an array, it only contains a single element. Therefore, I changed it from an array to a string.


aglensmith:
@traci707 did you test creating a token with an `allowed_cors_origins` set to a string? Typically, swapping out an array for a string won't work.

traci707:
@aglensmith Austin, you are correct. It did not work without the array. I will change it back but it will cause an error in the file.

aglensmith (approved):

PR #337, opened 2021-05-13
Remove origin_channel_id in Customers v3

capsula4:
Remove any mention of `origin_channel_id`. This is not a public concept and should not be mentioned.


reference/customers.v3.yml
```diff
@@ -248,11 +248,9 @@ paths:
         **Notes**
         For multi-storefront merchants who wants to give:
         * Global access:
-          * The channel listed in `origin_channel_id` needs `allow_global_logins=true` in their channel settings
-          * Leave `channel_ids` empty
-        * Channel restricted access:
-          * The channel listed in `origin_channel_id` needs to have `allow_global_logins=false` in their channel settings
-          *  Either leave `channel_ids` empty to use the current channel or give the list of channels we grant access
+          * Leave `channel_ids` empty. Default storefront channel should have the Channel Customer Setting `allow_global_logins=true`.
```
valentindellangela, line 251: 🤔  Do we explain somewhere that channel needs to have `allow_global_logins=true`?
I'm wondering if developers won't understand it as `to give global access, just leave channel_ids empty`

valentindellangela (approved):
My feedback is not blocking, just wondering if we can improve the sentence.

reference/customers.v3.yml
capsula4: Probably there's a gap in the documentation, as we haven't really explained anywhere in this section what a channel with global access is. I think it might be worth explaining further or at least adding a link.

valentindellangela: I remember recommending to Charles to write an article about this feature since it's quite complex and difficult to summarise just in the API doc.

valentindellangela (approved):

reference/customers.v3.yml
capsula4: I have added a separate note with the link to the Customer Setting Channel endpoint at least so they know where they can set/read that setting.

reference/customers.v3.yml
```diff
@@ -246,13 +246,13 @@ paths:
...
-        * Global access:
-          * The channel listed in `origin_channel_id` needs `allow_global_logins=true` in their channel settings
-          * Leave `channel_ids` empty
-        * Channel restricted access:
-          * The channel listed in `origin_channel_id` needs to have `allow_global_logins=false` in their channel settings
-          *  Either leave `channel_ids` empty to use the current channel or give the list of channels we grant access
+
+        A customer can be created with global access or channel-specific access.
+        * **Global access:**
+          * Make sure the default storefront has `allow_global_logins` enabled. This is on by default only for the default storefront. Find more info at [Customer Settings > Channel](https://developer.bigcommerce.com/api-reference/store-management/customers-v3/customer-settings-channel/customersettingschannelget)
```
valentindellangela, line 252: 🍹 
```suggestion
          * Make sure the storefront has `allow_global_logins` enabled. This is on by default only for the default storefront. Find more info at [Customer Settings > Channel](https://developer.bigcommerce.com/api-reference/store-management/customers-v3/customer-settings-channel/customersettingschannelget)
```

valentindellangela (approved):

reference/customers.v3.yml
```diff
@@ -246,13 +246,13 @@ paths:
...
-        * Channel restricted access:
-          * The channel listed in `origin_channel_id` needs to have `allow_global_logins=false` in their channel settings
-          *  Either leave `channel_ids` empty to use the current channel or give the list of channels we grant access
+
+        A customer can be created with global access or channel-specific access.
+        * **Global access:**
+          * Make sure the channel has `allow_global_logins` enabled. This is on by default only for the default storefront. Find more info at [Customer Settings > Channel](https://developer.bigcommerce.com/api-reference/store-management/customers-v3/customer-settings-channel/customersettingschannelget)
+          * Omit `channel_ids` field, or provide `channel_ids: null`.
+        * **Channel-specific access:**
+          * Provide an array `channel_ids` containing the channel ids that the customer should be granted access to.
```
traci707, line 255: * Provide a `channel_ids` array containing customer accessible `channel_ids`.

traci707 (commented):
I wanted to rephrase the following sentence so it doesn't end with a preposition.

Changed
Provide an array `channel_ids` containing the channel ids that the customer should be granted access to.
to
Provide a `channel_ids` array containing customer accessible `channel_ids`.

traci707 (approved):

PR #354, opened 2021-05-28
DEVDOCS-2821-carts.v3.yml

traci707:
[DEVDOCS-2821](https://jira.bigcommerce.com/browse/DEVDOCS-2821)

Added option_selections example and updated schema


traci707:
Valentine said 
"Unfortunately once merchants start using our API we cannot change the fields name as it will require them to update their app (for those using it) or break their site :confused:
I agree though this should have been more consistent!"

This means we can't change these fields so they are consistent
is_taxable and taxable
options and options_selections

aglensmith (approved):

PR #370, opened 2021-06-22
DEVDOCS-2888-inventory_level

traci707:
[DEVDOCS-2887](https://jira.bigcommerce.com/browse/DEVDOCS-2887) and [DEVDOCS-2888](https://jira.bigcommerce.com/browse/DEVDOCS-2888)

Added `0` for true, `1` for false to the following parameters:
`is_free_shipping`
`inventory_level`
`inventory_level:in`
`inventory_level:not_in`
`inventory_level:min`
`inventory_level:max`
`inventory_level:greater`
`inventory_level:less`
`inventory_low`

### DIFF
--- a/reference/catalog.v3.yml
+++ b/reference/catalog.v3.yml
@@ -207,8 +207,7 @@ paths:
           in: query
           type: boolean
         - name: is_featured
-          description: |
-            Filter items by is_featured. `0` for true, `1` for false.
+          description: 'Filter items by is_featured. `0` for false, `1` for true.'
           required: false
           in: query
           type: integer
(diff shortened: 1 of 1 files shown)


ektravel:
@traci707 Can you please ping me when this PR is ready for review?

traci707:
Oleksandr comments:

oleksandr.lysenko:computer:  3 days ago
@yuliia.budnikova  FYI. Yulia, Traci made a legit notes to Catalog API both documentation and API itself. Indeed 1) params is_featured and is_free_shipping are filtered by integers 1 or 0 not as the booleans as it is declared. (I think we could leave this as is)  2) description for is_featured says 0 for true, 1 for false which isn’t right (should be changed only in documentation) 3) inventory_level params filtered not only by integers but also float numbers.( I guess we doing automatical rounding to integer for input param since inventory could be only integer. I don’t think we should do anything both in documentation and API side)

PR #373, opened 2021-06-28
DEVDOCS-2817-sort-and-order-consistency

traci707:
[DEVDOCS-2817](https://jira.bigcommerce.com/browse/DEVDOCS-2817)

Updated the following endpoints:
[Get Redirects](https://developer.bigcommerce.com/api-reference/store-management/redirects/redirects/getredirects)
[Get All Scripts](https://developer.bigcommerce.com/api-reference/store-management/scripts/scripts/getscripts) 
to match 
[Get All Products](https://developer.bigcommerce.com/api-reference/store-management/catalog/products/getproducts)


reference/scripts.v3.yml
```diff
@@ -180,7 +180,8 @@ paths:
           in: query
           type: integer
         - name: sort
-          description: Field name to sort the scripts by.
+          description: |
+            Field name to sort the scripts by. Note: Since `id` increments when new scripts are added, you can use that field to sort by scripts create date.
```
markcmurphy, line 184:
```suggestion
            Field name to sort the scripts by. Note: Since `id` increments when new scripts are added, you can use that field to sort by script create date.
```

markcmurphy (approved):

PR #384, opened 2021-07-08
DEVDOCS-2927-update-required-fields

traci707:
[DEVDOCS-2927](https://jira.bigcommerce.com/browse/DEVDOCS-2927)

Tested the endpoint and the required attributes listed at the top need to be updated.

The required attributes should be the following:
Required Fields
* customer_id – ID of the Customer
* attribute_id – ID of the Customer Attribute
* value – Value of the Customer Attribute

[
  {
    "attribute_id": 2,
    "value": "Small",
    "customer_id": 19
  }
]


ektravel (approved):
LGTM

PR #393, opened 2021-07-12
DEVDOCS-2877 also it says id is req'd, but it's attribute_id that is req'd

markcmurphy:
[DEVDOCS-2877](https://jira.bigcommerce.com/browse/DEVDOCS-2877)

- didn't see where `id` was incorrectly marked as required like this ticket mentions but I did add a couple of examples and fixed a small typo.


reference/customers.v3.yml
```diff
@@ -4326,7 +4328,7 @@ components:
           type: integer
           format: int32
         value:
-          description: 'Attribute value. This will always be a string, regardless of the attributes type.'
+          description: 'Attribute value. This will always be a string, regardless of the attribute''s type.'
```
uwikaiddi, line 4331:
```suggestion
          description: 'Attribute value. This will always be a string, regardless of the attribute's type.'
```

uwikaiddi (approved):

PR #396, opened 2021-07-14
DEVDOCS-2931-consignment-discountid-checkouts.v3.yml

traci707:
[DEVDOCS-2931](https://jira.bigcommerce.com/browse/DEVDOCS-2931)

Changed consignment discount id from a string to an integer.


dcushing (approved):
Looks good to me!

markcmurphy (approved):

PR #398, opened 2021-07-15
DEVDOCS-2905-400-response

traci707:
[DEVDOCS-2905](https://jira.bigcommerce.com/browse/DEVDOCS-2905)
A 400 response has been added to various endpoints (see JIRA ticket above) to help users troubleshoot when they receive a 400 response.

        '400':
          description: |-
            Bad Request
            If the server allows only requests that match a limited list of common browser user agents, the server might return a generic 400 status code. For example, suppose your server sends a request to the Ruby UserAgent, and the Ruby UserAgent is not on the list of approved user agents. You will receive a 400 response unless you add it to the list.

### DIFF
--- a/reference/catalog.v3.yml
+++ b/reference/catalog.v3.yml
@@ -1405,6 +1405,13 @@ paths:
                 url_tiny: 'https://cdn8.bigcommerce.com/s-id30h7ohwf/products/158/images/485/product-image__98178.1536854227.66.100.png?c=2'
                 date_modified: '2018-09-13T15:57:07+00:00'
               meta: {}
+        '400':
+          description: |-
+            Bad Request. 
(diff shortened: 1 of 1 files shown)


markcmurphy:
I tested this out for a bit and tried to download an image from a server that blocks all requests. I got the following message back: 

```json
{
    "status": 400,
    "code": 400,
    "title": "The field 'image_file' is invalid.",
    "type": "https://developer.bigcommerce.com/api-docs/getting-started/api-status-codes",
    "errors": {
        "invalid_reason": "The resource at 'https://murphymark.me/wp-content/uploads/2021/01/womens-fashion-fit-t-shirt-caribbean-blue-60072dc7bd2b3__15063.1611083233.1280.1280.png' could not be downloaded and may be invalid."
    }
}
```

The `invalid_reason` returned makes sense for both bad requests and being blocked by firewall software. Rather than going into detail about the Ruby client, maybe we can just add the `400` with a description along the lines of: 

"Bad request. The requested resource could not be downloaded and may be invalid. Possible reasons include malformed request syntax or the file host blocking requests."

markcmurphy (approved):

PR #402, opened 2021-07-19
Devdocs 2640 update webhooks admin api documentation for open beta

markcmurphy:
[DEVDOCS-2640](https://jira.bigcommerce.com/browse/DEVDOCS-2640)

This involved merging the `api-specs` version of `webhooks.v3.yml` file with the version containing the beta Webhooks Admin endpoint. To compare them item by item, I changed the order of objects to match the OAS specs which accounts for most of the additions and deletions. I combined info or selected whichever was most comprehensive. For instance, the beta Webhooks Admin doc had some error codes for POST `/webhooks` that aren't in the `api-specs` repo.

I published a preview version that might help with reviewing:   

[Stoplight Front-end Preview](https://bc4wp1.docs.stoplight.io/api-reference/storefront/webhooks2)


reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
 openapi: 3.0.0
 info:
-  title: Webhooks v3
+  title: Webhooks
+  description: |
+    Create and manage webhooks to get notified when specific events occur on stores. For more information, [see Webhooks Overview](https://developer.bigcommerce.com/api-docs/store-management/webhooks/overview).
```
ektravel, line 5:
```suggestion
    Create and manage webhooks to get notified when specific events occur on stores. For more information, see [Webhooks Overview](https://developer.bigcommerce.com/api-docs/store-management/webhooks/overview).

```

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
 info:
-  title: Webhooks v3
+  title: Webhooks
+  description: |
+    Create and manage webhooks to get notified when specific events occur on stores. For more information, [see Webhooks Overview](https://developer.bigcommerce.com/api-docs/store-management/webhooks/overview).
+
+     <div class="HubBlock--callout">
+     <div class="CalloutBlock--info">
+     <div class="HubBlock-content">
+    <!-- type:info -->
```
ektravel, line 10:
```suggestion

```

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
+    </div>
   version: ''
-  description: 'Get notified when specific events occur on a BigCommerce store. For more information, see [Webhooks Overview](https://developer.bigcommerce.com/api-docs/store-management/webhooks/overview).'
+  contact: {}
+servers:
+  - url: 'https://api.bigcommerce.com/stores/{$$.env.store_hash}/v3'
+    variables:
+      $$.env.store_hash:
+        default: '{$$.env.store_hash}'
+  - url: 'https://api.bigcommerce.com/stores/{$$.env.store_hash}/v2'
```
ektravel, line 25: Can we have two serves referenced on the same page? I believe we should separate V2 and V3 APIs.

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
+      operationId: getAllWebhooks
+      summary: Get Webhooks
+      description: |-
+        Returns a list of all webhooks on a store associated to the `client_id` used to authenticate the request.
+
+        <div class="HubBlock--callout">
+        <div class="CalloutBlock--info">
+        <div class="HubBlock-content">
+
+        <!-- type:info -->
```
ektravel, line 41:
```suggestion
```

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
+
+        > ### Note
+        > * Webhooks endpoints are available on both the V2 and V3 REST API.
```
ektravel, line 44: Do we need to mention the V2 API in this documentation?

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+      tags:
+        - Webhooks
+    post:
+      operationId: createWebhooks
       summary: Create a Webhook
       description: Creates a webhook. Only one webhook at a time can be created. Custom headers can be added. Destination URL must be served on port 443 (custom ports are not currently supported).
```
ektravel, line 102:
```suggestion
      description: Creates a webhook. You can only create one webhook at a time. Custom headers are permitted. Destination URL must be served on port 443 (custom ports are not currently supported).

```

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
+        '400':
+          description: Bad request.
+        '401':
+          description: Unauthorized - the v3 Auth client ID or token in the request are not a valid combination for this store.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+        '404':
+          description: Webhook with specified id doesn't exist.
```
ektravel, line 228:
```suggestion
          description: Webhook with specified ID doesn't exist.

```

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
       tags:
         - Webhooks
     delete:
+      operationId: deleteAWebhook
+      summary: Delete a Webhook
+      description: 'Deletes a webhook. Only one webhook at a time can be deleted. When a webhook is deleted, it is returned in the response as a `200` OK.'
```
ektravel, line 244:
```suggestion
      description: 'Deletes a webhook. You can create only one webhook at a time. When a webhook is deleted, it is returned in the response as a `200` OK.'

```

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
+    get:
+      operationId: getHooksAdmin
+      summary: Get Admin Info
+      description: |-
+        List all notification emails, webhooks and blacklisted domains associated with the API user.
+
+        <div class="HubBlock--callout">
+        <div class="CalloutBlock--info">
+        <div class="HubBlock-content">
+        <!-- type:info -->
```
ektravel, line 289:
```suggestion

```

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
+
+        > ### Note
+        > * Webhooks endpoints are available on both the V2 and V3 REST API.
```
ektravel, line 292: Do we need to mention the V2 REST API in this documentation?

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
+          type: boolean
+        created_at:
+          description: UTC timestamp in seconds that the webhook was created.
+          type: integer
+          format: int64
+        updated_at:
+          description: UTC timestamp in seconds that the webhook was last updated.
+          type: integer
+          format: int64
+    WebhookCreate:
```
ektravel, line 528: I see you are changing the naming convention. Is there a reason for this update?

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,642 @@
...
           type: string
-          example: 'https://665b65a6.ngrok.io/webhooks`'
-          description: URL must be active, return a 200 response, and be served on port 443 (custom ports not currently supported).
-        is_active:
-          type: boolean
-          example: true
-          description: If webhook is active or not
+          format: url
         headers:
+          description: Headers to be included in destination url request.
```
ektravel, line 540:
```suggestion
          description: Headers to be included in destination URL request.

```

ektravel (commented):
I don't think we should be referencing the V2 REST API unless absolutely necessary. Similarly, I don't think we should list both servers (V2 and V3). When it comes to changing the naming convention, I will defer to @uwikaiddi.

reference/webhooks.v3.yml
markcmurphy: I left the mention since it was in the API beta docs. Let's see what @uwikaiddi says about whether we should break this spec file into two menu items for `V2` and `V3` or leave consolidated, in which case we should probably point out that both versions are available.

reference/webhooks.v3.yml
markcmurphy: Same as above about waiting to see if we are going to break this into two spec files for `V2` and `V3`.

uwikaiddi (approved):

markcmurphy:
Converting to draft because I have paused DEVDOCS-2640 until I receive a go-ahead from webhooks.

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,644 @@
...
+  description: |
+    Create and manage webhooks to get notified when specific events occur on stores. For more information, see [Webhooks Overview](https://developer.bigcommerce.com/api-docs/store-management/webhooks/overview).
+
+
+     <div class="HubBlock--callout">
+     <div class="CalloutBlock--info">
+     <div class="HubBlock-content">
+
+    > ### Note
+    > * Webhooks endpoints are available on both the V2 and V3 REST API.
```
jyuricle, line 13: We should take into account that V2 & V3 endpoints are a bit different. For example, `getAllWebhooks` endpoint for v2 doesn't support pagination. Also, in v2 response json is not wrapped in `data` parent field. So it might have sense to differentiate it until v2 is not fully removed.

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,644 @@
...
+          description: Identifies whether webhook is active.
+          type: boolean
+        created_at:
+          description: UTC timestamp in seconds that the webhook was created.
+          type: integer
+          format: int64
+        updated_at:
+          description: UTC timestamp in seconds that the webhook was last updated.
+          type: integer
+          format: int64
```
jyuricle, line 528: On the UI it shows that these fields have 3 validations and the interval for them is from -9223372... to 9223372. At least the minimum should be 0.

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,644 @@
...
+                        scope: '`store/order/*`'
+                        destination: 'https://665b65a6.ngrok.io/webhooks`'
+                        is_active: true
+                        headers:
+                          custom: string
+                    meta:
+                      pagination:
+                        offset: 0
+                        limit: 0
+                        total_items: 0
```
jyuricle, line 87: V3 pagination which is used for webhooks as well has different structure: 
```suggestion
                      pagination:
                        total: 0
                        count: 0
                        per_page: 0
                        current_page: 0
                        total_pages: 0
                        links:
                          previous: ?limit=0&page=0
                          current: ?limit=0&page=0
                          next: ?limit=0&page=0

```

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,644 @@
...
           properties:
             custom:
               type: string
+        is_active:
+          description: Identifies whether webhook is active.
+          type: boolean
       required:
         - scope
         - destination
         - is_active
```
jyuricle, line 555: Actually it is not required, but `is_active=true` by default

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,644 @@
...
-          type: string
-    put:
       responses:
         '200':
-          $ref: '#/components/responses/webhook_Resp'
+          description: Successful operation.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/WebhookResponse'
```
jyuricle, line 167: This schema misses such fields as `client_id`, `created_at`, `updated_at`, `store_hash`. It should be the same webhook schema as in `getAllWebhooks`

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,644 @@
...
+            is_active: true
+            scope: store/cart/lineItem/*
+        required: true
+      responses:
+        '200':
+          description: Webhook updated.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/WebhookResponse'
```
jyuricle, line 218: The same as in `Get webhook`, schema is not complete

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,644 @@
...
         '200':
-          $ref: '#/components/responses/webhook_Resp'
-      summary: Delete a Webhook
-      description: 'Deletes a webhook. Only one webhook at a time can be deleted. When a webhook is deleted, it is returned in the response as a 200 OK.'
-      operationId: deleteAWebhook
+          description: Webhook deleted.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/WebhookResponse'
```
jyuricle, line 257: The same as in `Get webhook`, schema is not complete

reference/webhooks.v3.yml
```diff
@@ -1,463 +1,644 @@
...
+                        type: array
+                        items:
+                          type: string
+                          format: email
+                      hooks_list:
+                        description: 'List of all the webhooks associated with the provider API User account, filtered by ''active'' parameter.'
+                        type: array
+                        items:
+                          $ref: '#/components/schemas/Webhook'
+                      blacklisted_domains:
```
jyuricle, line 326: This is represented in JSON a bit differently: 
```
    "blocked_domains": [
      {
        "destination": "string",
        "time_left": 0,
        "reasons": [
          {
            "failure_description": "string",
            "count": 0,
            "timestamp": 0
          }
        ]
      }
 ```

File: reference/webhooks.v3.yml [hunk 21]
```diff
@@ -1,463 +1,644 @@
...
+                    type: string
+                    format: email
+            examples:
+              Example:
+                value:
+                  emails:
+                    - user@example.com
+        required: true
+      responses:
+        '200':
```
jyuricle, line 428: Should 204 with `No content`

PR #408, opened 2021-07-22
Devdocs 2961 - Fix Content-Type on request runner for POST /catalog/products/{product_id}/images

markcmurphy:
[DEVDOCS-2961
](https://jira.bigcommerce.com/browse/DEVDOCS-2961)
Send a Test Request on https://developer.bigcommerce.com/api-reference/store-management/catalog/product-images/createproductimage defaults to `multipart/form-data`. Switching the order of the content types consumed fixed that. 

Also added `400` and `422` responses.


reference/catalog.v3.yml
```diff
@@ -1424,8 +1430,39 @@ paths:
...
+            required:
+              - status
+              - title
+              - type
+            x-examples:
+              example-1:
+                status: 422
+                title: image_url must be present if uploading by url
+                type: 'https://developer.bigcommerce.com/api-docs/getting-started/api-status-codes'
+          examples:
```
traci707, line 1459: Just wondering if the example below is necessary since we have the same example above.

traci707 (approved):
LGTM

PR #411, opened 2021-07-23
DEVDOCS-2974-discounted-lineItems

traci707:
[DEVDOCS-2974](https://jira.bigcommerce.com/browse/DEVDOCS-2974)
Added a note to the following endpoints to make merchants aware of the case when an item is added back to their cart after they deleted it:

Storefront
[Delete Cart Line Item](https://developer.bigcommerce.com/api-reference/storefront/carts/cart-items/deletecartlineitem)
[Delete a Line Item](https://developer.bigcommerce.com/api-reference/storefront/checkouts/checkout-cart-items/checkoutscartsitemsitemidbycheckoutidandcartiddelete)

Management API 
CartsV3
[Delete Cart Line Item](https://developer.bigcommerce.com/api-reference/store-management/carts/cart-items/deletecartlineitem)


ektravel (commented):
Does this apply to both Storefront and Management APIs? According to the ticket info, this mainly applies to merchants who use custom solutions instead of the BC's native cart or checkout. Do you think this should be noted?

traci707:
Yes, this applies to Storefront and Management APIs.

ektravel (approved):
Approving since this is specifically for edge cases.

PR #416, opened 2021-07-26
DEVDOCS-2977-added-royalmail

traci707:
[DEVDOCS-2977](https://jira.bigcommerce.com/browse/DEVDOCS-2977)

Added Royalmail to the Create Order Shipment endpoint definition.


bc-rogq (approved):

reference/orders.v2.oas2.yml
```diff
@@ -3979,7 +3535,7 @@ components:
           type: string
         customer_locale:
           type: string
-          example: \"en\", \"en-us\",
+          example: '\"en\", \"en-us\",'
```
markcmurphy, line 3538: I think this should just be a single string without the escape characters. The original swagger file was attempting to show multiple examples, but we should stick with a single example for the spec file. If I try to update an order with anything other than a single string such as `"en"`, `"en-au"`, or `"en-gb"`,  I get back the following error:

```json
  {
    "message": "The field 'customer_locale' is invalid.",
    "status": 400
  }
```

```suggestion
          example: 'en'
```

reference/orders.v2.oas2.yml
```diff
@@ -1534,10 +1526,10 @@ components:
                   street_2: ''
                   city: Austin
                   state: Texas
-                  zip: '78108'
+                  zip: 78108
                   country: United States
                   country_iso2: US
-                  phone: '1234567890'
+                  phone: 1234567890
```
markcmurphy, line 1532: Making a `GET` request to `/orders/{order_id}` returns both phone as a string when I test it.

markcmurphy (approved):

Issue #327, opened 2021-05-06
Catalog V3 spec issues


stevenbrookes:
Hi 

We saw a couple of issues in the V3 catalog specification whcih gave incorrect code generation.

1.  Additional  type: string on line 18199 productVariant_Full

2. On line 19182 product_Post
            $ref: '#/definitions/productVariant_Post_Product'
 should be 
            type: array
            items:
              $ref: '#/definitions/productVariant_Post_Product'

3.  On line 20026 product_Full
                   $ref: '#/definitions/productVariant_Full'
should be
            type: array
            items:
              $ref: '#/definitions/productVariant_Full'

Regards

Steve

stevenbrookes:
Hi

Another one for you the update product endpoint should take an array of productVariant_Put_Product, not just the object.

  product_Put:
    description: The model for a PUT to update a product.
    allOf:
      - type: object
        properties:
          id:
            type: integer
            description: |
              The unique numerical ID of the product; increments sequentially.
            x-required:
              - put
      - $ref: '#/definitions/product_Base'
      - type: object
        properties:
          variants:
            type: array
            items:
              $ref: '#/definitions/productVariant_Put_Product'


Is there much point me posting these fixes for you?

Regards

Steve

stevenbrookes:
Next one

Putting                   allOf:  when you only have one object is unnecessary and does cause issues with some client library builder.

Example

Around line 2187

     responses:
        '200':
          description: ''
          schema:
            type: object
            title: Variant Collection Response
            properties:
              data:
                type: array
                items:
                  allOf:
                    - $ref: '#/definitions/productVariant_Full'
              meta:
                $ref: '#/definitions/metaCollection_Full'

The allOf is unnecessary.

ektravel:
@stevenbrookes Thank you for bringing this to our attention.

traci707:
> Hi @stevenbrookes 

Thank you for bringing this information to our attention. I have been able to find every item on your list except for the first one. Is there a variable that should be also of string type? I am not understanding the first one.
> 
> 1. Additional  type: string on line 18199 productVariant_Full

stevenbrookes:
Hi Traci

Regarding   1.	Additional type: string on line 18199 productVariant_Full

I can see that issue has already been fixed, so you can ignore that one now.

uwikaiddi:
Hi @stevenbrookes,
This issue has been converted into an internal JIRA ticket. 
[...]

PR #418, opened 2021-07-28
Devdocs 2859 catalog issues

traci707:
[DEVDOCS-2859](https://jira.bigcommerce.com/browse/DEVDOCS-2859)

Fixed issues in GitHub issue #327. Other issues still remain but these changes removed some errors.

See screenshot in ticket confirming the need for an array.

### DIFF
--- a/reference/catalog.v3.yml
+++ b/reference/catalog.v3.yml
@@ -19339,6 +19339,8 @@ definitions:
         type: string
         description: The Manufacturer Part Number (MPN) for the variant.
   productVariant_Full:
+    type: array
+    items:
     allOf:
       - $ref: '#/definitions/productVariant_Base'
       - type: object
--- a/reference/catalog.v3.yml
+++ b/reference/catalog.v3.yml
@@ -2393,8 +2393,7 @@ paths:
               data:
                 type: array
                 items:
-                  allOf:
-                    - $ref: '#/definitions/productVariant_Full'
+                  - $ref: '#/definitions/productVariant_Full'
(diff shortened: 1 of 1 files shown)


reference/catalog.v3.yml
```diff
@@ -19339,6 +19339,8 @@ definitions:
         type: string
         description: The Manufacturer Part Number (MPN) for the variant.
   productVariant_Full:
+    type: array
```
ektravel, line 19342: I don't think productVariant_Full is an array. It is an object consisting of productVariant_Base and some additional properties.

reference/catalog.v3.yml
```diff
@@ -19594,6 +19596,8 @@ definitions:
   productVariant_Post_Product:
     description: |
       The model for a POST to create variants on a product.
+    type: array
```
ektravel, line 19599: The way product_Post is structured, I don't think we should be changing productVariant_Post_Product to an array.

reference/catalog.v3.yml
```diff
@@ -20369,7 +20373,9 @@ definitions:
       - type: object
         properties:
           variants:
-            $ref: '#/definitions/productVariant_Put_Product'
+            type: array
```
ektravel, line 20376: Can you please verify that productVariant_Put_Product should be an array? It appears to be an object.

ektravel (commented):
From looking at the Req and res schemas, I don't think we should be changing these objects to arrays.

PR #419, opened 2021-07-28
DEVDOCS-2955 The schema for the attributes object in the v3 Customers endpoint is not accurate

markcmurphy:
[DEVDOCS-2955](https://jira.bigcommerce.com/browse/DEVDOCS-2955)

Verified that the schema was incorrect. Updated file to match the `attributes` object returned when using the `?include=attributes` query param on a `GET` request to `/v3/customers`.


ektravel (approved):
LGTM

PR #424, opened 2021-07-30
Updated example

ektravel:
Updated example. We currently only support 2-letter lang code for default_shopper_language.


traci707 (approved):
LGTM

PR #426, opened 2021-07-30
DEVDOCS-2932 Update documentation for DELETE /settings/email-statuses

markcmurphy:
[DEVDOCS-2932](https://jira.bigcommerce.com/browse/DEVDOCS-2932)

Add specs from the updated `settings.v3.yml` file.


funivan (approved):

reference/settings.v3.yml
```diff
@@ -769,16 +769,58 @@ paths:
...
+                product_review_email:
+                  nullable: true
+            examples:
+              Reset gift certificate template:
+                value:
+                  gift_certificate_email: null
+              Reset product review and invoice email templates:
+                value:
+                  invoice_email: null
+                  product_review_email: null
```
traci707, line 823: description: Provide email types that should be deleted

traci707 (commented):
Looks good. If you like you can add the following line.

description: Provide email types that should be deleted

reference/settings.v3.yml
markcmurphy: @traci707 That is still listed in the description on line 773. Are you referring to another field?

markcmurphy: @traci707 the description for the `DELETE` method is on 773, is that what you're referring to?

traci707: There were two descriptions in the swagger file supplied by the engineer. It is optional.

traci707 (approved):

PR #446, opened 2021-08-08
docs(orders): ORDERS-4237 Add descpription and example for the order …

dornah:
### What 
Update the api documentation to explain the supported product option types in orders api.

### Why
In the [PR](https://github.com/bigcommerce/bigcommerce/pull/42121) for ORDERS-618 we are adding support for new modifier types (text, multi-line, numbers only and date). The order.products.product_options.value can accept different formats of data depending on the option type which requires description.

### Proof
Screen Shot 2021-08-09 at 8 03 23 am
Screen Shot 2021-08-09 at 8 02 43 am

@bigcommerce/orders


reference/orders.v2.oas2.yml
```diff
@@ -4295,6 +4305,17 @@ components:
...
+                description: |-
+                  Depending on the option type, value can be one of the following:
+                  - The variant option value id or the modifier value id for modifier types with a list of defined values, such as drop down or checkbox modifiers.
+                  - The modifier value for modifier types without a list of defined values, such as text field or date field modifiers.
+
+                  Notes:
+                  - The following option types are not currently supported via api:
+                    - File Upload
+                    - Pick List
+                  - For date modifiers use either the `ISO-8601` or the `YYYY-mm-dd` format. The date values are saved and retuned as timestamps. For values entered using the YYYY-mm-dd format the store timezone is used when creating the timestamp.
```
donald-nguyen-bc, line 4317: maybe we should be specific which format is supported as ISO-8601 allows quite a number of formats: 

image

reference/orders.v2.oas2.yml
```diff
@@ -4295,6 +4305,17 @@ components:
...
+                  - For mulit-line text field modifiers, use the `\n` characters for separating the lines.
```
donald-nguyen-bc, line 4318: typo "mulit-line" 😄 

Also we may need to mention about those multi-line text modifiers only support 2 lines of text

donald-nguyen-bc (approved):
👍

reference/orders.v2.oas2.yml
dornah: TIL, thanks 👍 
I thought ISO-8601 is the complete one with TZ.
Will update it.

reference/orders.v2.oas2.yml
dornah: The two line constraint, is only when they have set it themselves in catalog.

reference/orders.v2.oas2.yml
dornah: How is the updated version?

donald-nguyen-bc: looks good now 👍 , thanks Dorna

dornah: 🙇‍♀️

WillemHoman (approved):

reference/orders.v2.oas2.yml
```diff
@@ -4268,6 +4268,16 @@ components:
...
+      x-examples:
+        Product With a Drop Down and a Text Field Modifier:
+          value:
+            product_id: 100
+            quantity: 1
+            product_options:
+              - id: 111
+                value: '123'
+              - id: 112
+                value: Custom Message
```
traci707, line 4280: Is Custom Message a string? Should it be 'Custom Message' ?

reference/orders.v2.oas2.yml
```diff
@@ -4268,6 +4268,16 @@ components:
...
         - `xxx`, `xxx_customer`, and `xxx_merchant` default to the value from `catalog` if not supplied in the request.
       type: object
+      x-examples:
+        Product With a Drop Down and a Text Field Modifier:
+          value:
+            product_id: 100
+            quantity: 1
+            product_options:
+              - id: 111
+                value: '123'
```
traci707, line 4278: Is '123' a string or integer?

reference/orders.v2.oas2.yml
```diff
@@ -4295,6 +4305,17 @@ components:
...
               value:
                 type: string
+                description: |-
+                  Depending on the option type, value can be one of the following:
+                  - The variant option value id or the modifier value id for modifier types with a list of defined values, such as drop down or checkbox modifiers.
+                  - The modifier value for modifier types without a list of defined values, such as text field or date field modifiers.
+
+                  Notes:
+                  - The following option types are not currently supported via api:
+                    - File Upload
```
traci707, line 4315: - File upload

reference/orders.v2.oas2.yml
```diff
@@ -4295,6 +4305,17 @@ components:
...
+                    - Pick List
```
traci707, line 4316: - Pick list

reference/orders.v2.oas2.yml
```diff
@@ -4295,6 +4305,17 @@ components:
...
+                  - For date modifiers use either the `YYYY-MM-DDThh:mm:ss±hh:mm` or the `YYYY-MM-DD` ISO-8601 formats. The date field modifier values are saved and retuned as timestamps. For values entered using the YYYY-MM-DD format the store timezone is used when creating the timestamp.
```
traci707, line 4317: For date modifiers use either the `YYYY-MM-DDThh:mm:ss±hh:mm` or the `YYYY-MM-DD` ISO-8601 formats. The date field modifier values are saved and returned as timestamps. For values entered using the YYYY-MM-DD format, use the store timezone when creating the timestamp.

traci707 (changes requested):

reference/orders.v2.oas2.yml
dornah: It is a string. Will update, thanks.

reference/orders.v2.oas2.yml
dornah: In practice for this particular case both string and int will work, but since we are documenting this field as string I will leave it like this.

reference/orders.v2.oas2.yml
dornah: > For values entered using the YYYY-MM-DD format, use the store timezone when creating the timestamp.

We are doing the timestamp creation in the server. What I meant is that if they use the date only format we will use the store timezone to convert it to timestamp, but if they provide the timezone using the YYYY-MM-DDThh:mm:ss±hh:mm the provided timezone is respected.

traci707: Thanks Dornah, this text should be fine

traci707 (approved):

PR #448, opened 2021-08-10
DEVDOCS-3000-add-404

traci707:
Added 404 response for the Variant Option Values endpoint.


ektravel (approved):
Confirmed with @traci707 that empty object is correct.

PR #468, opened 2021-08-19
DEVDOCS-2935-update-option-value

traci707:
[DEVDOCS-2935](https://jira.bigcommerce.com/browse/DEVDOCS-2935)
There is a bug within the Carts > Add Cart Line Item endpoint. The checkout team doesn't have time to look into the issue so I would like to make the documentation as correct as possible until things are fixed. It was determined that the `option_value` field is not a string. See the following slack discussions:
dev-documentation
https://bigcommerce.slack.com/archives/C57SQ401K/p1625854941125300
api-support
https://bigcommerce.slack.com/archives/C071LJPGX/p1625851962376000

So we know the `option_value` should be an integer.  I don't know if the format of the integer should be in epoch time.  I have not been able to confirm this.  So for now, I thought I could at least change `option_value` to an integer.


ektravel (approved):
Approved. It would be good to mention that the value must be in unix timestamp format (once confirmed).

traci707:
Once I know if it is unix or epoch format, I will add this information.

PR #505, opened 2021-09-08
Devdocs 2463 update webhooks docs

markcmurphy:
[DEVDOCS-2463](https://jira.bigcommerce.com/browse/DEVDOCS-2463)

This started out as an attempt to add the fields that trigger webhook events into the existing docs. There wasn't a great way to do that since many webhooks shared scopes and descriptions were in tables, so I've combined the following docs into a single parent element in the `/api-reference` side nav:

https://developer.bigcommerce.com/api-reference/store-management/webhook-callbacks
https://developer.bigcommerce.com/api-docs/store-management/webhooks/events
https://developer.bigcommerce.com/api-reference/store-management/webhooks

Callbacks have individual pages so they can be linked to and easily discovered. In the future, I would like to update the descriptions to match the style guide and make the language more consistent. For now, those have been left mostly unchanged.


ektravel (commented):
Found a couple of things:
- remove X-Auth-Client from security requirements
- OpenAPI 3.0 component names (including schema names) can only contain the characters A-Z a-z 0-9 - . Having * in schema names throws a semantic error.

markcmurphy:
> Found a couple of things:
> 
> * remove X-Auth-Client from security requirements
> * OpenAPI 3.0 component names (including schema names) can only contain the characters A-Z a-z 0-9 - . Having * in schema names throws a semantic error.

Ok, I removed X-Auth-Client and changed the `*` to "wildcard"

ektravel (approved):
LGTM

PR #510, opened 2021-09-10
DEVDOCS-3082 Update webhooks with new params from swagger

markcmurphy:
[DEVDOCS-3082](https://jira.bigcommerce.com/browse/DEVDOCS-3082)

Added new query params to GET /hooks and updated the meta properties that are returned.


reference/webhooks.v3.yml
```diff
@@ -2725,15 +2806,33 @@ components:
...
-          description: Total number of items in the result set.
-        total_items:
+          description: Number of items on current page.
+        per_page:
           type: integer
-          description: Total number of items.
+          description: Maximum number of items per page.
+        current_page:
+          type: integer
+          description: Current page number of results.
```
ektravel, line 2820: This item description is a bit confusing. Does it pertain to the page number or the number of results per current page?

ektravel (commented):
Added one suggestion.

reference/webhooks.v3.yml
markcmurphy: Good call, that is confusing. I have just updated to "Current page number"

ektravel (approved):
Swagger Editor did not detect any errors.

PR #511, opened 2021-09-13
DEVDOCS-3011 Update POST Create a Customer Address info

markcmurphy:
[DEVDOCS-3011](https://jira.bigcommerce.com/browse/DEVDOCS-3011)

- Add a note to POST customers/addresses that body does not accept state or province codes
- update values in example to US address and fill in missing fields
- change the order of properties in 200 response to match the actual response format
- shorten example response body to only one customer instead of 5+
- remove required fields from POST description since info is also in the schema
- remove `required` flag from shared example since it doesn't apply
- create another shared response for customer address POST and PUT without meta, since that info is only present in GET responses

### DIFF
--- a/reference/customers.v3.yml
+++ b/reference/customers.v3.yml
@@ -521,18 +521,7 @@ paths:
       tags:
         - Customer Addresses
     post:
-      description: |-
-        Creates a Customer Address. Multiple customer addresses can be created in one call.
-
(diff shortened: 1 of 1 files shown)


traci707 (approved):
LGTM. Glad you did some cleanup while in this file.

PR #534, opened 2021-10-13
DEVDOCS-3165 - Update customer impersonation API docs to clarify usage

markcmurphy:
Many of the changes here are MD formatting differences from using stoplight instead of VS code to edit the file that shouldn't make any difference in the final.

Mostly just removed `X-Bc-Customer-Id: {{CUSTOMER_ID}}` from the Create Customer Impersonation Token request and clarified the language a bit.


traci707 (approved):
LGTM!

PR #547, opened 2021-11-05
DEVDOCS-3181 Update storefront order API documentation

markcmurphy:
- Changed `lineItems.phsicalItems.discounts` to be oneOf either an empty array or a discount object.
- Stoplight changed some double quotes to single


traci707 (approved):
LGTM

PR #565, opened 2021-11-24
revert mistaken push duplicating pending DEVDOCS 3275

slsriehl:
see title


bc-tgomez (approved):
Approving per @slsriehl 's request

PR #591, opened 2022-01-19
DEVDOCS 3388 Fix wrapping schema 

bc-andreadao:
### What was asked: 
Fix the wrapping schema in the tax estimate request and response schemas.

### Problem:  
There was a reference error in the request and response schemas (wrapping key references itself)

### Changes that were made: 
-Created a new model for request (`request-item-wrapping`) and a new model for response (`response-item-wrapping`) that contain the `wrapping` key. These 2 new models reference the old models (`request-item` and `response-item`). I deleted the `wrapping` key from the old models, which fixes the circular reference problem.  

-`Items` contains the wrapping key, but both `Shipping` and `Handling` objects do not contain the `wrapping` key (confirmed  this with the Sample Tax Provider API). Hence, `Shipping` and `Handling` still reference `request-item` and `response-item`, but `Items` now references the new `request-item-wrapping` and `response-item-wrapping` models. 

-The circular reference errors occurred in 3 endpoints (Estimate Taxes, Commit Tax Quote, and Adjust Tax Quote) because all 3 endpoints use the same models. Thus, fixing the models simultaneously fixed the circular reference errors for all 3 endpoints (all 3 endpoints were changed) 

-The example responses in the "Estimate Taxes" and "Commit Tax Quote" endpoint were changed to reflect the changes in the schemas.


reference/tax_provider.yml
```diff
@@ -173,80 +172,78 @@ paths:
...
-                                  amount: 225
-                                  tax_class:
-                                    class_id: '0'
-                                    name: Brutal Tax
-                                    code: US
-                                  id: Brutal Tax
-                            type: item
-                            wrapping: null
-                          - id: d2675662-6326-4a23-9107-ab71fa6a21a1
+                          - id: string
```
markcmurphy, line 180:
```suggestion
                          - id: array
```
Maybe this will work better!

PR #592, opened 2022-01-19
Devdocs 3407 legacy coupons not aware

bc-tgomez:
Updated documentation to include notes about the legacy coupon codes only working with the store's default currency.


bc-tgomez:
Closing PR to start new changes.

PR #598, opened 2022-01-24
DEVDOCS-3374 Added max character limit for staff_notes key

bc-andreadao:
[DEVDOCS-3374](https://jira.bigcommerce.com/browse/DEVDOCS-3374)

Added max character limit for staff_notes key


traci707 (approved):
I added the limit to the other staff_notes field.

PR #599, opened 2022-01-24
DEVDOCS-3375 Added max character limit for comments on order shipments

bc-andreadao:
[DEVDOCS-3375](https://jira.bigcommerce.com/browse/DEVDOCS-3375)

Added max character limit for comments on order shipments


traci707 (approved):
LGTM

PR #606, opened 2022-01-31
Update orders.v2.oas2.yml

bc-tgomez:
DEVDOCS-3421
- Updated the tax_provider_id.
- Made grammar edits to the article.


reference/orders.v2.oas2.yml
```diff
@@ -2777,11 +2777,11 @@ components:
                 example: a8458391-ef68-4fe5-9ec1-442e6a767364
                 type: string
               status:
-                description: 'The status will include one of the (string, optiona) - values defined under Order Statuses. This value is read-only. Do not attempt to modify or set this value in a POST or PUT operation.'
+                description: 'The status will include one of the (string, options) - values defined under Order Statuses. This value is read-only. Do not attempt to modify or set this value in a POST or PUT operation.'
                 example: Awaiting Fulfillment
                 type: string
               custom_status:
-                description: 'Contains the same (string, optiona) - value as the Order Statuses object’s `custom_label` property.'
+                description: 'Contains the same (string, optional) - value as the Order Statuses object's `custom_label` property.'
```
markcmurphy, line 2784:
```suggestion
                description: 'Contains the same (string, optional) - value as the Order Statuses object’s `custom_label` property.'
```

markcmurphy:
```suggestion
                description: 'Contains the same (string, optional) - value as the Order Statuses object''s `custom_label` property.'
```

reference/orders.v2.oas2.yml
```diff
@@ -3082,20 +3082,19 @@ components:
               tax_provider_id:
                 type: string
                 description: |
-                  BasicTaxProvider - Tax is set to manual.
+                  BasicTaxProvider - Tax is set to manual and order is created in the store.
```
slsriehl, line 3085: @bc-tgomez are we sure the order has to be created in the store here? I'm not privy to your research. Could be good to check with engineering if you haven't already.

reference/orders.v2.oas2.yml
```diff
@@ -3082,20 +3082,19 @@ components:
...
 
-                  AvaTaxProvider - This is for when the tax provider has been set to automatic and the order was NOT created by the API. Used for Avalara.
+                  AvaTaxProvider - Tax is set to automatic and order is created in the store. Used for Avalara.
```
slsriehl, line 3087: Seems like the tax provider is Avalara and the tax provider was automatically set on the order (probably by the Avalara integration). Would another integration literally use this string, get a null because it used the API, or would it get its own new string?. It's an awkward set of things to say together, but IMO worth digging because of all the new tax provider services, like TaxJar.

reference/orders.v2.oas2.yml
```diff
@@ -3082,20 +3082,19 @@ components:
...
 
-                  "" (blank) - When the tax provider is unknown. This includes legacy orders and orders previously created via API.
-                  This can be set when creating an order using a POST.
+                  "" (null) - When the order is created by API, or the tax provider is unknown.
```
slsriehl, line 3089:
```suggestion
                  "" (null) - The order is created by API, or the tax provider is unknown.
```

reference/orders.v2.oas2.yml
```diff
@@ -4223,12 +4222,11 @@ components:
           type: string
         tax_provider_id:
           description: |
-            BasicTaxProvider - Tax is set to manual.
+            BasicTaxProvider - Tax is set to manual and order is created in the store.
```
slsriehl, line 4225: Same comment as previous `tax_provider_id: "BasicTaxProvider"`.

reference/orders.v2.oas2.yml
```diff
@@ -4223,12 +4222,11 @@ components:
...
 
-            AvaTaxProvider - This is for when the tax provider has been set to automatic and the order was NOT created by the API. Used for Avalara.
+            AvaTaxProvider - Tax is set to automatic and order is created in the store. Used for Avalara.
 
-            "" (blank) - When the tax provider is unknown. This includes legacy orders and orders previously created via API.
-            This can be set when creating an order using a POST.
+            "" (Null) - When the order is created with the API, or the tax provider is unknown.
           enum:
             - BasicTaxProvider
             - AvaTaxProvider
```
slsriehl, line 4232:
```suggestion
            - AvaTaxProvider
            - ''
```
enum should include empty string

reference/orders.v2.oas2.yml
```diff
@@ -4447,7 +4445,7 @@ components:
           example: false
           type: boolean
         order_source:
-          description: Orders submitted via the store's website will include a `www` value. Orders submitted via the API will be set to `external`. A read-only value. Do not pass in a POST or PUT.
+          description: Orders submitted from the store’s website will include a `www` value. Orders submitted with the API will be set to `external`. A read-only value. Do not pass in a POST or PUT.
```
slsriehl, line 4448:
```suggestion
          description: Orders submitted from the store’s website will include a `www` value. Orders submitted with the API will be set to `external`. This value is read-only. Do not pass in a POST or PUT.
```
Parallel to your previous edit.

reference/orders.v2.oas2.yml
bc-tgomez: @WillemHoman does this statement --what Sarah said-- still stand here? Would another integration use the string `AvaTaxProvider`?

reference/orders.v2.oas2.yml
bc-tgomez: From what I had discussed with [Willem](https://bigcommerce.slack.com/archives/C96P5T3J4/p1643313600099500) that was the rule. It populates at order creation by any means except `/v2/orders`.

bc-tgomez:
> I went on a spree replacing `via` and changing straight possessive apostrophes to curly ones, and I have suggestions you can take or leave.
> 
> I also have a question about the automatic tax provider enum `AvaTaxProvider` -- how does that play with other tax providers like TaxJar? Do they just set to '', or did engineering add to the enum list? If they set to '', probably should be clarified for their developers.

Im tagging @WillemHoman in the comment because from what I remember it was correct --but now that I'm revisiting our convo I'm not so sure. 

--Willem would you mind clarifying this again? and Thank you for your eyes on this. :)

reference/orders.v2.oas2.yml
WillemHoman: 👍 that's correct

reference/orders.v2.oas2.yml
WillemHoman: This will be tax provider specific. So Avalara will be `AvaTaxProvider`, Basic Tax is `BasicTaxProvider`. 
This list will expand over time.
I just [pinged the tax team](https://bigcommerce.slack.com/archives/C97B8DEGK/p1646864901260169) on this and there is no publicly available list of tax providers and tax provider ids. 
Apparently it also varies by region, so this sounds like something that should be documented for tax and linked to from  the orders api docs?
Would a couple of examples suffice for now?

bc-tgomez: That would be great to add for clarity. Thanks so much @WillemHoman.

WillemHoman:
> I also have a question about the automatic tax provider enum AvaTaxProvider -- how does that play with other tax providers like TaxJar? Do they just set to '', or did engineering add to the enum list? If they set to '', probably should be clarified for their developers.

`tax_provider_id` is not writable by `/v2/orders`, it always ends up as `''`, it should really be moved into the `readonly` fields and respond with an error message if merchants try to set it. 

I've checked with the tax team previously and they are adamant that it should not be publicly writable. 

If an order is placed via s2s checkout api, storefront or CP manual order it will have the appropriate Tax Provider id set automatically. 

It is pulled from [this](https://github.com/bigcommerce/bigcommerce/blob/ac4772f91ee71484ed541128254620f9a777bda4/app/Tax/Provider/Configuration/TaxProviderRegistry.php#L18) giant `enum` which gets updated whenever a new tax provider is added. 

So TaxJar will have a `tax_provider_id` of `TaxJarProvider`,  Digital River: `DigitalRiverProvider` etc

PR #593, opened 2022-01-19
SHIPPING-2200 Add delete pickup methods spec

bc-luke:
Add Open API documentation for [new delete pickup methods operation](https://github.com/bigcommerce/api-proxy-java/pull/1544).

⚠️ This API is not generally available, so the documentation shouldn't be made public yet. It'll first be part of a beta.

@bigcommerce/dev-docs

PR #607, opened 2022-02-01
SHIPPING-2201 Add Update pickup methods operation

agchan12:
This is branched of https://github.com/bigcommerce/api-specs/pull/593

DON'T MERGE UNLESS https://github.com/bigcommerce/api-specs/pull/593 has been merged.

Add Open API documentation for new update pickup methods operation.

⚠️ This API is not generally available, so the documentation shouldn't be made public yet. It'll first be part of a beta.

@bigcommerce/dev-docs


reference/pickup.v3.yaml
```diff
@@ -0,0 +1,99 @@
+openapi: 3.0.1
+info:
+  title: Pickup
```
bc-andreadao, line 3:
```suggestion
  title: Pickup Methods
```

reference/pickup.v3.yaml
```diff
@@ -0,0 +1,99 @@
...
+      store_hash:
+        default: store_hash
+paths:
+  '/stores/{store_hash}/v3/pickup/methods':
+    delete:
+      tags:
+        - Pickup Methods
+      description: Delete existing pickup methods.
+      parameters:
+        - name: id:in
```
bc-andreadao, line 18: id?

reference/pickup.v3.yaml
```diff
@@ -0,0 +1,99 @@
...
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/PickupMethod'
+                  meta:
+                    type: object
+        '422':
+          description: The request contains invalid data and may only partially be fulfilled. More information can be found in the error description.
+components:
+  schemas:
+    PickupMethod:
```
bc-andreadao, line 70: List which properties are required

bc-andreadao (changes requested):
Commented on a couple of points to clarify

reference/pickup.v3.yaml
agchan12: There may be future plans to expand upon this domain such that it's still under the `pickup` domain but not necessarily relating to pickup methods.

reference/pickup.v3.yaml
agchan12: I was thinking that this can be kept as the base model as this can be used for responses (which don't have a required property). I've updated the api-spec and have included the base model specifically in the request here https://github.com/bigcommerce/api-specs/pull/607/files#diff-941a81c827338c2ca7d86849e24d30cb8c5057711d9ae9f7e0032050d21e9a9cR31-R35

reference/pickup.v3.yaml
bc-luke: It actually is `id:in`, since it's a comma separated list of IDs, so we're saying find those with ID _in_ this list, for example [in customers API](https://github.com/bigcommerce/api-specs/blob/bdf5c3e41746c2636065c3bf17e18759ef5834c7/reference/customers.v3.yml#L90).

slsriehl:
Corollary devdocs ticket [DEVDOCS-3541](https://jira.bigcommerce.com/browse/DEVDOCS-3541).

cc @bc-andreadao

agchan12:
Closing this PR because it has already been merged in here https://github.com/bigcommerce/api-beta-multi-location-inventory/pull/7

agchan12:
This PR has been merged in here https://github.com/bigcommerce/api-beta-multi-location-inventory/pull/7

PR #617, opened 2022-02-10
DEVDOCS-3430 Added missing Auth Scope

bc-andreadao:
[DEVDOCS-3430](https://jira.bigcommerce.com/browse/DEVDOCS-3430)

Added missing auth scopes to Custom Template Associations


traci707 (approved):
LGTM!

Issue #582, opened 2022-01-06
Swagger 2.0 Schema Violations


dleguellecfortuitas:
Catalog_v3.yml does not parse properly in https://editor.swagger.io/ nor via NSwagStudio
![Example](https://user-images.githubusercontent.com/38509772/148468069-7c329e0a-23d6-4fac-a772-8d280e08960c.PNG)

el-besto:
A few other key collections for our integration project are broken outside of Stoplight:
- __Orders V3__ - violates with the use of `anyOf` on [line 2173](https://github.com/bigcommerce/api-specs/blob/7f99c30ba81369f53174b269c7993d1c86c1bdae/reference/orders.v3.yml#L2173) - perhaps upgrade to openapi 3.0.0?
- __Customers V2__ - needs to update its security definition [line 2900-2901](https://github.com/bigcommerce/api-specs/blob/7f99c30ba81369f53174b269c7993d1c86c1bdae/reference/customers.v2.yml#L2900)
- __Price List V2__ (see the violation error messages below)

## Customers V2 errors
```
Semantic error at security.1
Security requirements must match a security definition
Jump to line 2901
```

## Price List Errors
```
Structural error at info.contact.url
should match format "uri"
format: uri
Jump to line 83

Semantic error at paths./pricelists.get.parameters.8
Schemas with 'type: array', require a sibling 'items: ' field
Jump to line 146

Semantic error at paths./pricelists.get.parameters.9
Schemas with 'type: array', require a sibling 'items: ' field
Jump to line 149

Semantic error at paths./pricelists/{price_list_id}/records.get.parameters.16
Schemas with 'type: array', require a sibling 'items: ' field
Jump to line 903

Semantic error at paths./pricelists/{price_list_id}/records.get.parameters.17
Schemas with 'type: array', require a sibling 'items: ' field
Jump to line 906

Semantic error at paths./pricelists/assignments.get.parameters.4
Schemas with 'type: array', require a sibling 'items: ' field
Jump to line 2431

Semantic error at paths./pricelists/assignments.get.parameters.5
Schemas with 'type: array', require a sibling 'items: ' field
Jump to line 2435

Semantic error at paths./pricelists/assignments.get.parameters.6
Schemas with 'type: array', require a sibling 'items: ' field
Jump to line 2439

Semantic error at paths./pricelists/assignments.get.parameters.7
Schemas with 'type: array', require a sibling 'items: ' field
Jump to line 2443

[...]

PR #628, opened 2022-02-22
Devdocs 3442 schema violations

bc-andreadao:
[DEVDOCS-3442](https://jira.bigcommerce.com/browse/DEVDOCS-3442)

I posted screenshots of **updated** schema violations in the ticket ([DEVDOCS-3442](https://jira.bigcommerce.com/browse/DEVDOCS-3442)). Because Stoplight supports both openapi 2.0 and 3.0, I did not correct schema violations that occurred in Swagger due to problems with openapi version. However, each schema violation is explained below and in issue #582 :

**Schema violations that were corrected here** (i.e. problems that violated both openapi 2.0 and 3.0 formatting) include:
-Added "/" to basepath when path starts at host root, rather than leaving basepath empty (openapi 3.0 lets you leave the field empty to default to "/", but the term basepath implies 2.0 formatting). See [reference for 2.0](https://swagger.io/docs/specification/2-0/api-host-and-base-path/) and [reference for 3.0](https://swagger.io/docs/specification/api-host-and-base-path/)
-Added `items` field to `array` types. The `items` field includes they `type` for items in the array. This is required in openapi formatting. See [reference](https://swagger.io/docs/specification/data-models/data-types/?sbsearch=Data%20Types)
-Fixed the formatting for multiple data types (adding oneOf before listing the 2 types). See [reference](https://swagger.io/docs/specification/data-models/data-types/?sbsearch=Data%20Types)

**Schema violations in Swagger that did not need to be corrected in Stoplight**:
-"url" formatting is "uri" violation: Urls in openapi 2.0 have the uri format which does not accept empty strings. However, urls in openapi 3.0 have the uri-reference format, in which empty strings are accepted. See [reference](https://www.rfc-editor.org/rfc/rfc3986#section-3)
-`anyOf` is valid in openapi 3.0, which Stoplight supports
-"no content as an additional property" violation: In openapi 3.0, responses can have `content` as an additional property. See [reference](https://swagger.io/specification/?sbsearch=content)

I opened new tickets to investigate changes for some schema violations that involved security requirements. These include:
-[DEVDOCS-3551](https://jira.bigcommerce.com/browse/DEVDOCS-3551) 
-[DEVDOCS-3552](https://jira.bigcommerce.com/browse/DEVDOCS-3552)


traci707 (approved):
LGTM!  Thank you for taking care of all these errors in the files.  This cleanup needed to be done. Back in October 2020, I removed many references to OAuth Client since it was no longer required for most APIs.

PR #633, opened 2022-02-24
DEVDOCS-3455 pages v3 lives again

slsriehl:
minimum viable documentation; need URLs ASAP.


reference/pages.v3.yml
```diff
@@ -426,13 +995,16 @@ components:
               type: string
               description: 'Where the page’s type is a contact form: contact email address that receives messages sent via the form. Must be unique.'
               maxLength: 255
+              default: ''''''
```
slsriehl, line 998:
```suggestion
              default: ''

```

markcmurphy (approved):
could use some fine-tuning, but approving to generate links ✅

PR #635, opened 2022-02-24
Devdocs 3527 internal flags

bc-andreadao:
[DEVDOCS-3527](https://jira.bigcommerce.com/browse/DEVDOCS-3527)

Added internal flags to models so that "schema" tab won't show up on frontend


traci707 (approved):
LGTM!

PR #641, opened 2022-03-01
DEVDOCS-3598: docs(checkout): CHECKOUT-6388 Add Address field as an alias to ShippingAddress in the Consignment API Request

bc-mackxu:
## What?
Creating an alias to shippingAddress called Address in both Storefront and S2S Api request and deprecate (do not delete) the shippingAddress field.

## Why?
As discussed, it would make more sense to return an Address field in the consignment api so it works for Pickup and Shipping - Given this change, we think it also makes sense to update the request to match the response.

## Testing / Proof
image
image

Revert

ping @bigcommerce/checkout  @bigcommerce/dev-docs

### DIFF
--- a/reference/checkouts.sf.yml
+++ b/reference/checkouts.sf.yml
@@ -1,4 +1,4 @@
-swagger: '2.0'
+swagger: '3.0'
 info:
   version: ''
   title: Storefront Checkouts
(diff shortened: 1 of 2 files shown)


traci707 (approved):
LGTM

slsriehl:
@traci707 let's wait for @markcmurphy to get back.  this PR updates these spec files to oas3, so we'll either have to revert or do conversion checks.

reference/checkouts.sf.yml
```diff
@@ -1,4 +1,4 @@
-swagger: '2.0'
+swagger: '3.0'
```
bc-shawnwang, line 1: Should we update to 3.0 ? I saw many other files they use `openapi: 3.0.1`

When I tried to open it in the https://editor.swagger.io/ editor, it remind that: "Please indicate a valid Swagger or OpenAPI version field. Supported version fields are swagger: "2.0" and those that match openapi: 3.0.n (for example, openapi: 3.0.0)."

Also my VSCode Extension Swagger Preview cannot work for this change.

bc-mackxu: @slsriehl @traci707 , It seems ` swagger: "2.0"` doesn't support `deprecated: true`, could we upgrade the document to `openapi: 3.0.1`?

reference/checkouts.sf.yml
```diff
@@ -3431,6 +3431,62 @@ definitions:
     type: object
     properties:
       shippingAddress:
+        deprecated: true
```
valentindellangela, line 3434: 👍

valentindellangela (approved):
LGTM - make sure your changes reach T3 before we merge the document update.

bc-mackxu:
Closed as part of https://github.com/bigcommerce/api-specs/pull/664

PR #642, opened 2022-03-02
DEVDOCS-3283-allow-webhook-schemas

traci707:
Display webhook schemas


bc-andreadao:
Good for testing x-internal flag

slsriehl (changes requested):
can we fix 1320 and 1375 switching? this is going to change the links.

Issue #542, opened 2021-10-25
Management API - v2 order schema for Multi-Currency-enabled stores


martin-hudak:
Hi everybody and thank you for your work :)

I've noticed that in management API v2/orders there are optional order fields that are present when the store has multiple currencies:

> The following additional fields are returned on orders when Multi-Currency is enabled on the store:
> store_default_currency_code - the currency code of the store’s default currency.
> store_default_to_transactional_exchange_rate - the exchange rate between the store’s default currency and the transactional currency used in the order.
> 

(https://developer.bigcommerce.com/api-reference/store-management/orders)

In orders.v2.oas2.yml these fields are present in response example, but not in the `#/components/schemas/order_Resp`.
Is there some technical problem or it just was forgotten?

Thank you,
Martin

slsriehl:
Many thanks, @martin-hudak .  I've opened a ticket for us to look into this.

slsriehl:
Closing this out. See changes in the following:
* #644 

Please open another issue if something else comes up. Thanks for participating!

PR #644, opened 2022-03-02
DEVDOCS-3441 Update orders.v2.oas2.yml

bc-tgomez:
Added feedback from [request](https://github.com/bigcommerce/api-specs/issues/542) and included the following:
- 'default_currency_code'
- 'store_default_currency_code'
- 'store_default_to_transactional_exchange_rate'
- Quick clean up on misspelled words.


reference/orders.v2.oas2.yml
```diff
@@ -4435,6 +4435,18 @@ components:
           description: The transactional currency ID. A read-only value. Do not pass in a POST or PUT.
           example: 1
           type: integer
+        default_currency_code:     
+          description: The currency code of the store's default currency; returned on orders when Multi-Currency is enabled on the store.
```
WillemHoman, line 4439: This actually the transactional currency code. 

These are not easy to keep straight. 
Here are my notes. 
- `currency_id` - the display currency id. Depending on the currency selected, the value may be different from the
  transactional currency. not in the api though. In v2 display currency is set to the transactional currency
  i.e. `default_currency_id`
- `currency_code` - the currency code of the display currency used to present prices to the shopper on the storefront.
  Depending on the currency selected, the value may be different from the transactional currency.
- `currency_exchange_rate` - the exchange rate between the store's default currency and the display currency. For orders
  created using the V2 endpoints, this value is always 1 (only in the storefront this value can be different to 1).
- `default_currency_id` - the transactional currency id.
- `default_currency_code` - the currency code of the transactional currency the shopper pays in.
- `store_default_currency_code` - the currency code of the store's default currency.
- `store_default_to_transactional_exchange_rate` - the exchange rate between the store's default currency and the  transactional currency used in the order. 

The default currency refers to the transactional currency which is the  currency the customer pays in. 
The display currency refers to the presentational currency used to present prices to  the shopper on the storefront.

N.B. if the display currency is transactional then the display currency columns will be populated with the transactional currency values

```mysql
select ordcurrencyid           as display_currency_id,                  -- v2 currency_id, defaults to tx currency id f not set
       -- display_currency_code    -- v2 currency_code is not stored but looked up from display_currency_id
       ordcurrencyexchangerate as store_default_to_display_exchange_rate, -- v2 currency_exchange_rate
       orddefaultcurrencyid    as transactional_currency_id,        -- v2 default_currency_id
       currency_code           as transactional_currency_code,      -- v2 default_currency_code
       store_default_currency_code,                                               -- v2 store_default_currency_code
       store_default_to_transactional_exchange_rate                  -- v2   store_default_to_transactional_exchange_rate
from orders
where orderid >= 1008;
```

reference/orders.v2.oas2.yml
```diff
@@ -4435,6 +4435,18 @@ components:
...
+          example: EUR
+          type: string
+        store_default_currency_code: 
+         description: The currency code of the store's default currency; returned on orders when Multi-Currency is enabled on the store.
```
WillemHoman, line 4443: This is returned regardless of whether Multi-Currency is enabled

reference/orders.v2.oas2.yml
```diff
@@ -4435,6 +4435,18 @@ components:
...
+         example: USD
+         type: string
+        store_default_to_transactional_exchange_rate:
+          description: The exchane rate between the store's default currency and the transactional currency used in the order; returned on orders when Multi-Currency is enabled on the store.
```
WillemHoman, line 4447: The currency code of the store's default currency; returned on orders when Multi-Currency is enabled on the store.
Will b e `1` when the store default currency is the same as the transactional currency.

WillemHoman (approved):
:clap: Thanks for the update

PR #670, opened 2022-03-15
feat(customers): CUST-1545 add documentation for form field support for addresses

mjhmatt:
We are supporting the sending of form field values for addresses when creating a customer this is to document that

image


traci707 (approved):
LGTM

PR #672, opened 2022-03-16
DEVDOCS-3625- add-get-order-response-body

traci707:
[DEVDOCS-3625](https://jira.bigcommerce.com/browse/DEVDOCS-3625)
Changed x-internal to false so the response body displays.


slsriehl:
fixed by #681

PR #674, opened 2022-03-21
DEVDOCS-3278

traci707:
[DEVDOCS-3278](https://jira.bigcommerce.com/browse/DEVDOCS-3278)
Removed `product-Id`


slsriehl:
@traci707 not sure why several of the `type: object` key-values are removed here.  can you shed some light?

traci707:
@slsriehl I have no idea. I did not make the changes. Stoplight made some automatic changes with `type: object`

PR #656, opened 2022-03-08
DEVDOCS-3569 Corrected type of method called on endpoint

bc-andreadao:
[DEVDOCS-3569](https://jira.bigcommerce.com/browse/DEVDOCS-3569)

-Corrected description of request (changed GET to POST)

PR #675, opened 2022-03-21
Corrected type of method called on endpoint (#656)

markcmurphy:
Co-authored-by: Mark Murphy <mark.murphy@bigcommerce.com>


slsriehl (approved):
check

PR #684, opened 2022-03-24
Devdocs 3658 remove delete modifier image api

traci707:
[DEVDOCS-3658](https://jira.bigcommerce.com/browse/DEVDOCS-3658)
Removed delete modifier image api endpoint

delete:
      description: |-
        Deletes a *Modifier Image*.

        Deletes the image that was set to show when the modifier value is selected.
      operationId: deleteModifierImage
      responses:
        '204':
          description: ''
      summary: Delete Modifier Image
      parameters:
        - in: header
          type: string
          name: Accept
          default: application/json
          required: true
        - name: Content-Type
          in: header
          type: string
          default: application/json
          required: true
      tags:
        - Product Modifier Images

### DIFF
--- a/reference/catalog.v3.yml
+++ b/reference/catalog.v3.yml
@@ -8227,29 +8227,6 @@ paths:
       summary: Create Modifier Image
       tags:
         - Product Modifier Images
-    delete:
-      description: |-
-        Deletes a *Modifier Image*.
-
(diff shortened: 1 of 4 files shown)


slsriehl:
@traci707 there are significant changes to this spec file. I'm closing this & reopening with your changes added different.

PR #700, opened 2022-03-30
Devdocs 3602 get prices store hash

bc-andreadao:
[DEVDOCS-3602](https://jira.bigcommerce.com/browse/DEVDOCS-3602)

Add `store_hash` as a field for making test requests


slsriehl (approved):
LGTM

PR #569, opened 2021-12-03
change-currency-DEVDOCS-3302

traci707:
[DEVDOCS-3302](https://jira.bigcommerce.com/browse/DEVDOCS-3302)
added cart/change-currency to Storefront Cart API

PR #710, opened 2022-04-03
DEVDOCS-3302?: docs(checkout): CHECKOUT-6553 Add documentation for storefront cart currency endpoint

bc-mackxu:
## What?
As above

## Why?
We created a while ago an endpoint `/cart/change-currency` to allow changing the currency of a cart on Storefront. However, we added it as a private endpoint so we didn't add the `storefront` namespace at the time.

Due to feedback from sales engineers and dev doc teams, it turns out developers would like to have public access to this endpoint in order to build custom solutions on the storefront that allows the shopper to change currency. 

In order to make this endpoint publicly available, we need to ensure the URL of the endpoint is consistent with other endpoints.

## Testing / Proof
image

Revert

ping @bigcommerce/checkout  @bigcommerce/dev-docs @traci707

### DIFF
--- a/reference/carts.sf.yml
+++ b/reference/carts.sf.yml
@@ -250,6 +250,45 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/deleteCartsItems'
+  '/cart/{cartId}/currency':
+    post:
+      tags:
+        - Cart Currency
+      summary: Update Cart currency
+      description: |-
+        Update currency of the *Cart*. 
+        Promotions and gift certificates that don't apply to the new currency will be removed from your cart.
+        The cart currency would not be able to update if the draft order cart or the cart contains manual discount.
+
+        <!-- theme: info -->
+        > #### Note
+        > The Send a Test Request feature is not currently supported for this endpoint.
+      operationId: addCartCurrency
+      parameters:
+        - name: cartId
+          in: path
+          description: This cart's unique ID.
+          required: true
+          schema:
+            type: string
+            format: UUID
+      responses:
+        '200':
+          $ref: '#/components/responses/getCarts'
+      requestBody:
+        content:
+          application/json:
+            schema:
+              title: Cart Currency Request Data
+              required:
+                - currencyCode
+              type: object
+              properties:
+                currencyCode:
+                  type: string
+                  description: currency code
+              x-internal: false
+        description: ''
 components:
   schemas:
     responseCartCoupons:


reference/carts.sf.yml
```diff
@@ -250,6 +250,43 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/deleteCartsItems'
+  '/cart/{cartId}/currency':
+    post:
+      tags:
+        - Cart Currency
+      summary: Update Cart currency
+      description: |-
+        Update currency of the *Cart*.
```
valentindellangela, line 259: Maybe we can add some notes around the fact that any promotions, gift certificate applied to the cart...will may get removed from the cart.

I think we send this message in Cornerstone for instance
image

valentindellangela (approved):

bc-shawnwang (approved):

reference/carts.sf.yml
bc-mackxu: Thank you for suggestion. Yes, we should add those description.

slsriehl:
@traci707 following the tickets, it looks like this is for [DEVDOCS-3302](https://jira.bigcommerce.com/browse/DEVDOCS-3302). I recall you have #569 open for this, and it looks like there's another branch, as well.

reference/carts.sf.yml
```diff
@@ -250,6 +250,45 @@ paths:
...
           $ref: '#/components/responses/deleteCartsItems'
+  '/cart/{cartId}/currency':
+    post:
+      tags:
+        - Cart Currency
+      summary: Update Cart currency
+      description: |-
+        Update currency of the *Cart*. 
+        Promotions and gift certificates that don't apply to the new currency will be removed from your cart.
+        The cart currency would not be able to update if the draft order cart or the cart contains manual discount.
```
slsriehl, line 261: The verbs in this sentence need some attention.

reference/carts.sf.yml
```diff
@@ -250,6 +250,45 @@ paths:
...
+      operationId: addCartCurrency
+      parameters:
+        - name: cartId
+          in: path
+          description: This cart's unique ID.
+          required: true
+          schema:
+            type: string
+            format: UUID
+      responses:
```
slsriehl, line 275: document error codes, please.

slsriehl:
@traci707 pull in some of your changes from #569 . You had some good stuff there, like a currency code example string in requestCartCurrency and an improved description of responseCartCurrency.

traci707:
@bc-mackxu From PR #569 , remove line 497 and replace with the following:

PR #714, opened 2022-04-05
DEVDOCS-3302 cart change currency

slsriehl:


reference/carts.sf.yml
```diff
@@ -250,6 +250,45 @@ paths:
...
           $ref: '#/components/responses/deleteCartsItems'
+  '/cart/{cartId}/currency':
+    post:
+      tags:
+        - Cart Currency
+      summary: Update Cart currency
+      description: |-
+        Update currency of the *Cart*. 
+        Promotions and gift certificates that don't apply to the new currency will be removed from your cart.
+        The cart currency would not be able to update if the draft order cart or the cart contains manual discount.
```
slsriehl, line 261: revise verbs

reference/carts.sf.yml
```diff
@@ -250,6 +250,45 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/deleteCartsItems'
+  '/cart/{cartId}/currency':
```
markcmurphy, line 253: @bc-mackxu This seemed to work with both `cart` and `carts`. Can we use `carts` here so it is consistent with the other endpoints? Results for `carts` pictured
 
image

bc-mackxu: Thank you @markcmurphy, yes, please update and consistent with other endpoints.

slsriehl: @traci707 FYI ^^

traci707:
Thank you @markcmurphy and @slsriehl, I think everything looks good. The only concern I have is the response.  @bc-mackxu has the response as  $ref: '#/components/responses/getCarts' which shows the id, email, currency, etc.  This is not what the output looks like as shown in Mark's screenshot unless I am missing something.

reference/carts.sf.yml
```diff
@@ -250,6 +250,45 @@ paths:
...
+      operationId: addCartCurrency
+      parameters:
+        - name: cartId
+          in: path
+          description: This cart's unique ID.
+          required: true
+          schema:
+            type: string
+            format: UUID
+      responses:
```
slsriehl, line 275: @traci707 any headway on documenting error response codes?

traci707: @slsriehl thank you for reminding me.

markcmurphy (approved):

PR #723, opened 2022-04-07
Devdocs 3784 add value to list

traci707:
# [DEVDOCS-3784](https://jira.bigcommerce.com/browse/DEVDOCS-3784)

## What changed?
`value`  cannot be edited unless using the owner's credentials, therefore added it with the others.

### DIFF
--- a/reference/catalog.v3.yml
+++ b/reference/catalog.v3.yml
@@ -12356,7 +12356,7 @@ paths:
       tags:
         - Product Metafields
       summary: Update a Product Metafield
-      description: "Updates a *Product Metafield*.\n\n**Required Fields**\n* none\n\n**Read-Only Fields**\n* id\n* These fields can only be modified by the app (API credentials) that created the metafield:\n\t* `namespace`\n\t* `key`\n\t* `permission_set`\n\t* `value`\n\n**Usage Notes**\n* Attempting to modify `namespace`, `key`, `permission_set`, and `value` fields using a client ID different from the one used to create those metafields will result in a `403` error message. "
+      description: "Updates a *Product Metafield*.\n\n**Required Fields**\n* none\n\n**Read-Only Fields**\n* id\n* These fields can only be modified using the API account that created the metafield:\n\t* `namespace`\n\t* `key`\n\t* `permission_set`\n\t* `value`\n\n**Usage Notes**\n* Attempting to modify the `namespace`, `key`, `permission_set`, or `value` field using an API account different from the one used to create those metafields will result in a `403` error message. "
       operationId: updateProductMetafield
       parameters:
         - name: metafield_id
--- a/reference/catalog.v3.yml
+++ b/reference/catalog.v3.yml
@@ -12356,7 +12356,7 @@ paths:
       tags:
         - Product Metafields
       summary: Update a Product Metafield
-      description: "Updates a *Product Metafield*.\n\n**Required Fields**\n* none\n\n**Read-Only Fields**\n* id\n* These fields can only be modified by the app (API credentials) that created the metafield:\n\t* namespace\n\t* key\n\t* permission_set\n\t* value\n\n**Usage Notes**\n* Attempting to modify `namespace`, `key`, `permission_set`, and `value` fields using a client ID different from the one used to create those metafields will result in a 403 error message. "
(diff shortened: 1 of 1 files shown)


reference/catalog.v3.yml
```diff
@@ -12356,7 +12356,7 @@ paths:
       tags:
         - Product Metafields
       summary: Update a Product Metafield
-      description: "Updates a *Product Metafield*.\n\n**Required Fields**\n* none\n\n**Read-Only Fields**\n* id\n* These fields can only be modified by the app (API credentials) that created the metafield:\n\t* namespace\n\t* key\n\t* permission_set\n\n**Usage Notes**\n* Attempting to modify `namespace`, `key`, and `permission_set` fields using a client ID different from the one used to create those metafields will result in a 403 error message. "
+      description: "Updates a *Product Metafield*.\n\n**Required Fields**\n* none\n\n**Read-Only Fields**\n* id\n* These fields can only be modified by the app (API credentials) that created the metafield:\n\t* namespace\n\t* key\n\t* permission_set\n\t* value\n\n**Usage Notes**\n* Attempting to modify `namespace`, `key`, `permission_set`, and `value` fields using a client ID different from the one used to create those metafields will result in a 403 error message. "
```
markcmurphy, line 12359:
```suggestion
      description: "Updates a *Product Metafield*.\n\n**Required Fields**\n* none\n\n**Read-Only Fields**\n* id\n* These fields can only be modified by the app (API credentials) that created the metafield:\n\t* `namespace`\n\t* `key`\n\t* `permission_set`\n\t* `value`\n\n**Usage Notes**\n* Attempting to modify `namespace`, `key`, `permission_set`, and `value` fields using a client ID different from the one used to create those metafields will result in a `403` error message. "
```
Code formatting

markcmurphy (changes requested):

reference/catalog.v3.yml
```diff
@@ -12356,7 +12356,7 @@ paths:
       tags:
         - Product Metafields
       summary: Update a Product Metafield
-      description: "Updates a *Product Metafield*.\n\n**Required Fields**\n* none\n\n**Read-Only Fields**\n* id\n* These fields can only be modified by the app (API credentials) that created the metafield:\n\t* namespace\n\t* key\n\t* permission_set\n\n**Usage Notes**\n* Attempting to modify `namespace`, `key`, and `permission_set` fields using a client ID different from the one used to create those metafields will result in a 403 error message. "
+      description: "Updates a *Product Metafield*.\n\n**Required Fields**\n* none\n\n**Read-Only Fields**\n* id\n* These fields can only be modified by the app (API credentials) that created the metafield:\n\t* `namespace`\n\t* `key`\n\t* `permission_set`\n\t* `value`\n\n**Usage Notes**\n* Attempting to modify `namespace`, `key`, `permission_set`, and `value` fields using a client ID different from the one used to create those metafields will result in a `403` error message. "
```
slsriehl, line 12359:
```suggestion
      description: "Updates a *Product Metafield*.\n\n**Required Fields**\n* none\n\n**Read-Only Fields**\n* id\n* These fields can only be modified using the API account that created the metafield:\n\t* `namespace`\n\t* `key`\n\t* `permission_set`\n\t* `value`\n\n**Usage Notes**\n* Attempting to modify the `namespace`, `key`, `permission_set`, or `value` field using an API account different from the one used to create those metafields will result in a `403` error message. "
```

slsriehl: the ambiguity here is that the credential most often passed in the API request is the `access_token`. For store API credentials, the `access_token` doesn't change. For app API credentials, the `access_token` can change, but the client ID doesn't change. Because the client ID will remain the same, the app can still successfully use this endpoint with a new `access_token`. But it's confusing because the client ID doesn't actually get sent in the API request.

traci707: Thanks for catching this @slsriehl. I guess this is an older doc that still contained `Client ID`.

slsriehl (approved):
LGTM

bobbyshaw:
Hi @traci707. I noticed this change while looking for something else. I believe the value field can be written to by other API accounts if the permission set allows for it (write or write with storefront access). So I wonder if the new description is a little misleading in that it doesn’t have quite the same restrictions as the key and namespace fields.

PR #730, opened 2022-04-11
DEVDOCS-3781: add-webdav-example

traci707:
# [DEVDOCS-3781](https://jira.bigcommerce.com/browse/DEVDOCS-3781)

## What changed?
Added an example using the webdav: prefix

### DIFF
--- a/reference/checkouts.v3.yml
+++ b/reference/checkouts.v3.yml
@@ -2363,7 +2363,7 @@ responses:
           order_confirmation_use_custom_checkout_script: false
           custom_order_confirmation_script_url: 'https://example.com/custom-order-confirmation-script.js'
         meta: {}
-      'webdav: prefix':
+      'WebDAV protocol':
         data:
           custom_checkout_script_url: 'webdav:vtz-checkout/dist/auto-loader.js'
           order_confirmation_use_custom_checkout_script: false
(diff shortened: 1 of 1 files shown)


reference/checkouts.v3.yml
```diff
@@ -2363,3 +2363,9 @@ responses:
           order_confirmation_use_custom_checkout_script: false
           custom_order_confirmation_script_url: 'https://example.com/custom-order-confirmation-script.js'
         meta: {}
+      'webdav: prefix':
```
slsriehl, line 2366:
```suggestion
      'webdav protocol':
```

slsriehl: `webdav:` is a protocol scheme -- [read more](https://en.wikipedia.org/wiki/WebDAV) if you wish

slsriehl: I'm not sure where the value of this YAML key shows up or what is signifies on the rendered page, so my suggestion might be in the wrong format.

valentindellangela: That's correct Wevdav is a protocol. Also FYI we haven't updated the code yet to allow webdav to be used. This will be implemented likely next sprint as part of https://jira.bigcommerce.com/browse/CHECKOUT-6576

traci707: @valentindellangela Thank you for this feedback.  I accepted the changes and will wait to merge this PR. Please let me know when it is in the code and ready for use.

slsriehl (approved):
approved, LGTM. Added awaiting-css-cue label.

FYI @traci707 cc @bc-keith

PR #732, opened 2022-04-12
DEVDOCS-3762: add-discount-to-checkout

traci707:
# [DEVDOCS-3762](https://jira.bigcommerce.com/browse/DEVDOCS-3762)

## What changed?
Added new endpoint, v3/checkouts/{checkoutId}/discounts

### DIFF
--- a/reference/checkouts.v3.yml
+++ b/reference/checkouts.v3.yml
@@ -2,7 +2,10 @@ swagger: '2.0'
 info:
   version: '3.0'
   title: Checkouts
-  description: Create checkouts from existing carts using BigCommerce checkout logic.
+  description: |-
+    For more information on authenticating BigCommerce APIs, see [Authentication](/api-docs/getting-started/authentication).
+
+    Create and manage checkouts from existing carts using BigCommerce checkout logic.
 host: api.bigcommerce.com
 basePath: null
 schemes:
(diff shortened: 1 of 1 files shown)


reference/checkouts.v3.yml
```diff
@@ -145,6 +145,69 @@ paths:
...
+                    items:
+                      type: object
+                      properties:
+                        discounted_amount:
+                          type: number
+                          example: 10
+                        name:
+                          type: string
+                          example: manual
+                        id:
```
valentindellangela, line 189: This api only accepts `discounted_amount` and `name` in the payload.

reference/checkouts.v3.yml
```diff
@@ -145,6 +145,69 @@ paths:
...
+        in: path
+        required: true
+    post:
+      summary: Add Discount to Checkout
+      operationId: post-store_hash-v3-checkouts-checkoutId-discounts
+      responses:
+        '200':
+          $ref: '#/responses/CheckoutResponse'
+      description: |-
+        Adds a discount to an existing *checkout*.
```
valentindellangela, line 165: Maybe we can add that this discount only applies to `line_items`?

traci707:
@valentindellangela Can you look at my changes?

reference/checkouts.v3.yml
```diff
@@ -1747,33 +1810,10 @@ parameters:
     enum:
       - consignments.available_shipping_options
 securityDefinitions:
-  X-Auth-Token:
+  Checkouts:
+    name: API Key
```
slsriehl, line 1814: based on the endpoints, this looks like a store-management API. if that's true, the OAuth stuff needs to go back in, and the access_token gets passed as the X-Auth-Token.

```suggestion
    name: X-Auth-Token
```

slsriehl (changes requested):
looks like the auth section may need to be cleaned up. depending on the rest of the file, this may be duplication that's correctly removed. please confirm and take appropriate action, if any.

@traci707

markcmurphy (approved):

PR #737, opened 2022-04-19
DEVDOCS-3729 - find and replace $$.env.store_hash with store_hash

markcmurphy:
# [DEVDOCS-3729](https://jira.bigcommerce.com/browse/DEVDOCS-3729)

## What changed?
* find and replace $$.env.store_hash with store_hash

### DIFF
--- a/reference/webhooks.v3.yml
+++ b/reference/webhooks.v3.yml
@@ -135,12 +135,12 @@ x-stoplight:
 servers:
   - url: 'https://api.bigcommerce.com'
     variables:
-      $$.env.store_hash:
-        default: '{$$.env.store_hash}'
+      store_hash:
+        default: '{store_hash}'
   - url: 'https://api.bigcommerce.com'
     variables:
-      $$.env.store_hash:
-        default: '{$$.env.store_hash}'
+      store_hash:
+        default: '{store_hash}'
 components:
   parameters:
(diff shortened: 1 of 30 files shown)


reference/webhooks.v3.yml
```diff
@@ -135,12 +135,12 @@ x-stoplight:
 servers:
   - url: 'https://api.bigcommerce.com'
     variables:
-      $$.env.store_hash:
-        default: '{$$.env.store_hash}'
+      store_hash:
```
slsriehl, line 138: on a couple different docs, i see these server variables. since we have them listed as path parameters, is it correct to keep them listed as server variables?

if we're keeping `variables: store_hash`, standardize across files. some have a `default` value of `{store_hash}` and others say `unknown`.

slsriehl:
Not all of these comments require action.

- Looks like all the instances changed should be changed. I am trusting that you found all the instances that should be changed -- I didn't check the latter.
- See comment about server variables.
- I can approve Wednesday. Alternatively, once you address the server variables bit, I am cool for you to roll this PR into a publish batch yourself while I'm out.

@markcmurphy

reference/webhooks.v3.yml
markcmurphy: Good call

PR #741, opened 2022-04-21
DEVDOCS-3836b: scope creep

slsriehl:
# [DEVDOCS-3836](https://jira.bigcommerce.com/browse/DEVDOCS-3836)B: Scope creep

## What changed?
* VS Code's version of Spectral is happy.

## Review questions
* I didn't log any of the requests -- are the new values in fact correct/plausible? ie, do we really get empty string instead of null, are the ID strings realistic, etc


slsriehl:
Okay, cool. It seems like swagger 2 maybe doesn't use `nullable: true`. Not sure how it allows nulls.

slsriehl:
on hold to await conversion to OAS3+

PR #743, opened 2022-04-21
DEVDOCS-3762b: Suggestions

markcmurphy:
I had a few suggestions for DEVDOCS-3762 and rather than make direct changes or bulky suggestions, I created another branch.

- Removed the "required fields" part of the description since that info is present in the body model
- Linked to some shared parameters to make updating easier in the future
- Added some descriptions to `name` and `discounted_amount` 
- Set minimum on `discounted_amount` since a negative amount returns an error
- Changed `discounted_amount` example to show decimal places
- Removed the example since it uses the example values from the schema, so Stoplight will display one automatically. This helps keeping the schema and example synced up.

### DIFF
--- a/reference/catalog.v3.yml
+++ b/reference/catalog.v3.yml
@@ -4153,7 +4153,21 @@ paths:
       tags:
         - Product Variants Metafields
       summary: Update Product Variant Metafields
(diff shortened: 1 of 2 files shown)


slsriehl:
@markcmurphy I resolved conflicts and converted inline descriptions to multi-line. No comment on the content.

@traci707 -- how do you feel? ready to codeowner approve on your say-so.

traci707:
@markcmurphy thank you for converting to multi-line, I did not know that caused the errors.  
@slsriehl I am fine with it.

PR #750, opened 2022-04-28
DEVDOCS-3881: fix(customers): CUST-2101 Adding form field schema to address under C…

bc-tgomez:
# [DEVDOCS-3881](https://jira.bigcommerce.com/browse/DEVDOCS-3881)

## Roll over PR to DevDoc branch
* Add changes to branch


reference/customers.v3.yml
```diff
@@ -3359,6 +3359,94 @@ components:
...
+        phone: '707070707'
+        postal_code: '33333'
+        state_or_province: California
+        form_fields: 
+          - name: "test"
+            value: "test"
+      type: object
+      properties:
+        first_name:
+          description: The first name of the customer address.
```
slsriehl, line 3382: These are not quite right... the name doesn't belong to the address. but it also might not directly belong to the customer? could be a gift recipient, for example. Reflect & consider revising.

reference/customers.v3.yml
```diff
@@ -3359,6 +3359,94 @@ components:
...
+        address2:
+          description: The address 2 line.
+          type: string
+        city:
+          description: The city of the customer address.
+          type: string
+          minLength: 0
+          maxLength: 100
+        state_or_province:
+          description: The state or province name. It is required for countries that need a state/province to complete an address.
```
slsriehl, line 3408:
```suggestion
          description: The state or province name. Required for countries that need a state/province to complete an address.
```

reference/customers.v3.yml
```diff
@@ -3359,6 +3359,94 @@ components:
...
+          type: string
+          minLength: 0
+          maxLength: 100
+        postal_code:
+          description: The postal code of the customer address. It is required for countries that need postal codes to complete an address.
```
slsriehl, line 3413:
```suggestion
          description: The postal code of the customer address. Required for countries in which valid addresses must have postal codes.
```

reference/customers.v3.yml
bc-tgomez: @bc-akshatkaushik can the description be reworded to better reflect the object type?

bc-tgomez:
Thanks @slsriehl! Good catches. I tagged @bc-akshatkaushik to see if we can better describe the customer address, i'd give it a go but I don't want to lose that meaning for the devs.

reference/customers.v3.yml
bc-akshatkaushik: @slsriehl So the address first name is separate to the customer first name, for example I as a customer can have an address with my friend's details if I want to have something delivered to him.
Do you still think a revision is required?

slsriehl: @bc-akshatkaushik this is helpful, thanks. It's not so much that a revision is required; it's that one of our value adds as a team is to optimize language to leave the reader with no feeling of loose ends.

@bc-tgomez and also Akshat, how do you feel about something like the following?

```suggestion
          description: The first name associated with the customer address.
```

bc-akshatkaushik: I agree @slsriehl that is definitely very important and something we tend to miss out on as developers.
Your suggested change is more in line with the internal implementation and intention of use for the field, looks good to me.
Thank You for pointing it out, looking forward to more of the same as we are in the process of planning to audit the customers devdocs space for inconsistencies with what is implemented.

markcmurphy (approved):

PR #751, opened 2022-04-28
DEVDOCS-3873-add-clarification-to-process-payment-endpoint

traci707:
# [DEVDOCS-3873](https://jira.bigcommerce.com/browse/DEVDOCS-3873)

## What changed?
Changed the URL and added more clarification regarding the value for Accept.


slsriehl (approved):
great within the scope of the ticket. 

If I'm not mistaken, the `payments` endpoint does not need the X-Auth-Token header; just the `Authorization: PAT jkahsdfk` header. Please make a follow-up ticket to investigate unless that was 100% covered in the work for this PR. As this appears now, it looks like `payments` is requiring X-Auth-Token.

traci707:
@slsriehl You are correct. We did discuss this and I should make this change. I will create a ticket.

PR #755, opened 2022-05-02
DEVDOCS-3894-removed-X-Auth-Token-from-process-payment-endpoint

traci707:
# [DEVDOCS-3894](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-3894)

## What changed?
Removed X-Auth-Token


reference/payment_processing.yml
```diff
@@ -317,6 +317,7 @@ paths:
                   - status
                   - title
                   - type
+      security: []
```
slsriehl, line 320: shouldn't this diff show something removed? or is this empty array overriding a security param at a higher level?

and what about the `Authorization: PAT {{pat_token}}` header? does that not count as security?

slsriehl: let's let it go for now. Make a low-priority ticket to follow up with payments.

traci707: Resolving conversation, see [DEVDOCS-4003](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4003)

PR #757, opened 2022-05-02
DEVDOCS-3884-increase-limits-to-250

traci707:
# [DEVDOCS-3884](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-3884)

## What changed?
 Increased metafields limits to 250 for the other resources. See the list in the JIRA ticket.
Resolved file errors

### DIFF
--- a/reference/catalog.v3.yml
+++ b/reference/catalog.v3.yml
@@ -3816,7 +3816,7 @@ paths:
         **Read-Only Fields**
         * id
 
-        **Note:** The max number of metafields allowed on each product, category, variant, or brand is 250 per client ID. For more information, see [Platform Limits](https://support.bigcommerce.com/s/article/Platform-Limits) in the Help Center.
+        **Note:** The maxiumum number of metafields allowed on each order, product, category, variant, or brand is 250 per client ID. For more information, see [Platform Limits](https://support.bigcommerce.com/s/article/Platform-Limits) in the Help Center.
       operationId: createVariantMetafield
       parameters:
         - name: product_id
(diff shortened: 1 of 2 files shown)


bookernath (dismissed):

reference/catalog.v3.yml
```diff
@@ -3816,7 +3816,7 @@ paths:
         **Read-Only Fields**
         * id
 
-        **Note:** The max number of metafields allowed on each product, category, variant, or brand is 250 per client ID. For more information, see [Platform Limits](https://support.bigcommerce.com/s/article/Platform-Limits) in the Help Center.
+        **Note:** The maxiumum number of metafields allowed on each order, product, category, variant, or brand is 250 per client ID. For more information, see [Platform Limits](https://support.bigcommerce.com/s/article/Platform-Limits) in the Help Center.
```
slsriehl, line 3819: @traci707 @bookernath so a product, etc, updated by multiple API keys could have (250 * x) metafields, where x is the number of API keys in the mix? Seems weird.

bookernath: Correct. The limit is per-entity-per-client-ID. This is an area where I think it's best that clients don't have to worry about the activity of other clients messing up their ability to use the system - it would be pretty frustrating if I have an app that relies on metafields, and I get some error related to "you can't add metafields to this product because other clients already added too many".

Metafields are pretty simple (key/value storage/retrieval) and the endpoints are paginated, so we're not too worried about it so far.

markcmurphy (approved):

PR #767, opened 2022-05-06
DEVDOCS-3985-update-variantId-and-productId

traci707:
# [DEVDOCS-3985](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-3985)

## What changed?
Changed double to integer

### DIFF
--- a/reference/checkouts.sf.yml
+++ b/reference/checkouts.sf.yml
@@ -2822,11 +2822,11 @@ definitions:
                           variantId:
                             description: ID of the variant.
                             type: number
-                            format: double
+                            format: integer
                           productId:
                             description: ID of the product.
                             type: number
-                            format: double
+                            format: integer
                           sku:
                             description: SKU of the variant.
                             type: string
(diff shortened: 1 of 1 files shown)


bc-tgomez (approved):
Looks good to go.

markcmurphy (dismissed):

reference/checkouts.sf.yml
```diff
@@ -3762,11 +3762,11 @@ definitions:
                   variantId:
                     description: ID of the variant.
                     type: number
-                    format: double
+                    format: integer
```
markcmurphy, line 3765: Rather than leaving the type as `number` and changing `format` to `integer`, these should instead be `type: integer` since `number` and `integer` are two distinct data types: https://swagger.io/docs/specification/data-models/data-types/

``` yaml
                  variantId:
                    description: ID of the variant.
                    type: integer
```

markcmurphy (changes requested):

markcmurphy (approved):

PR #779, opened 2022-05-16
DEVDOCS-4057 shipping links

bc-andreadao:
# [DEVDOCS-4057](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4057)

Fixed broken links

There's no landing page for each endpoint, so I linked an example method for the endpoint. 
preview


bc-andreadao:
> @bc-andreadao, the style guide recommends using "for example" instead of e.g. to make things as clear as possible.
> 
> A thought: instead of linking to examples, what about linking to the Shipping V2 landing page (https://developer.bigcommerce.com/api-reference/14c24a5d787b6-shipping-v2) which contains a brief overview of the three endpoints?

Good idea, changed the link to the landing page

markcmurphy (approved):

PR #785, opened 2022-05-18
DEVDOCS-4105: Use the unexploded form type for comma-separated string lists

S48-Mo:
## What changed?

* The definitions for `include`, `include_fields`, and `exclude` query params are changed to a comma-separated array of strings, instead of literal strings. This helps with machine-parsing and automated building of comma-separated lists of sub-resources to include and exclude as part of the ReST API calls. As an example: `"custom_fields,variants"` is considered to be a list of strings with enumerated values `custom_fields` and `variants`, rather than a single atomic string.

* A few drive-by changes to add a missing `data` field in some API responses, and some spelling corrections.


slsriehl:
Thanks, @S48-Mo!  I've created an internal ticket to run this by P & E. We appreciate y'all's participation!

slsriehl:
Merging to ticket branch for reconciliation with significantly different main branch tree.

PR #795, opened 2022-05-24
DEVDOCS-3720 Missing Product Video

bc-andreadao:
# [DEVDOCS-3720](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-3720)

Fixed example not showing up in docs (youtube links don't show up there because of stoplight)


reference/catalog.v3.yml
```diff
@@ -2673,17 +2673,17 @@ paths:
...
                     video_id:
+                      type: string
                       maxLength: 25
                       minLength: 0
-                      type: string
                       description: |
                         The ID of the video on a host site.
                       x-required:
                         - post
+                      example: 'https://www.youtube.com/watch?v=`R12345677`'
```
traci707, line 2685: The video is no longer available.

traci707 (commented):
@bc-andreadao Can you find a different video to use in the example?

bc-andreadao:
Note: `video_id` needed to be under 25, so I changed up the example

reference/catalog.v3.yml
bc-andreadao: I'm having trouble finding a video that's under 25 characters

bc-andreadao: Is it supposed to be a string of numbers?

traci707: @bc-andreadao oh okay, I think I see.

reference/catalog.v3.yml
```diff
@@ -2649,7 +2649,7 @@ paths:
                 The model for a POST to create a video on a product.
               allOf:
                 - title: Product Video Base
-                  type: object
+                  description: Common ProductVideo properties.
```
markcmurphy, line 2652:
```suggestion
                  description: Common Product Video properties.
```

reference/catalog.v3.yml
markcmurphy: @bc-andreadao The `video_id` is the string that comes after the `v` query param in the YouTube link. So if the video URL is `https://www.youtube.com/watch?v=z3fRu9pkuXE`, one would just need to enter `z3fRu9pkuXE`.

Here is a full request body:

```json
{
  "title": "Writing Great Documentation",
  "description": "A video about documenation",
  "sort_order": 1,
  "type": "youtube",
  "video_id": "z3fRu9pkuXE"
}
```

markcmurphy (changes requested):

reference/catalog.v3.yml
bc-andreadao: made the suggested change

reference/catalog.v3.yml
bc-andreadao: thanks, added in the example

markcmurphy (approved):
✅✅✅

PR #797, opened 2022-05-25
Update-create-a-product-modifier-endpoint

traci707:
# [DEVDOCS-4107](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4107)

## What changed?
Added clarifying text for a checkbox modifier with option values


reference/catalog.v3.yml
```diff
@@ -6893,6 +6893,9 @@ paths:
 
         **Read-Only Fields**
         * `id`
+
+        **Notes**
+        It takes two separate requests to create a new checkbox modifier with option values. First, perform a request to create a modifier. Then perform a second request to update option values.
```
bc-tgomez, line 6898: I just have a slight suggestion that you can take or leave. 😺 

"Then, perform a second request to update option values." or "Perform a request to create a modifier, then perform a second request to update option values."

bc-tgomez (approved):
Looks good! Made a suggestion to line 6898.

reference/catalog.v3.yml
traci707: Thank you @bc-tgomez 
Your suggestion is better. I should not have a first without a second.

markcmurphy (approved):

PR #816, opened 2022-06-09
docs(promotions): PROMO-404 add documentation of bulk deletion of cou…

bc-vincent-zhao:
…pon codes

# [DEVDOCS-4179](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4179)

## What changed?
* remove the wrong statement that we delete one page of data by default
* add optional query parameter `?id:in=1,2,3` to allow bulk deletion of coupons

ping @bigcommerce/orders


bc-vincent-zhao:
@bigcommerce/dev-docs would you please review this ? thanks 🙇

reference/marketing.v2.yml
```diff
@@ -222,8 +222,6 @@ paths:
         - Coupons
     delete:
       description: |
-        By default, it deletes a page of *Coupons*.
```
bc-vincent-zhao, line 225: this is not true, I've tested it and it deletes all coupons.

donald-nguyen-bc: do you think it's quite risky to have the delete all endpoints ? if merchant has 1 million of coupons and they call this API without the `id:in` param then the consequence will be huge as we have no way to recover.

bc-vincent-zhao: completely agree. This is as dangerous as a `DELETE table` SQL statement without the `where` clause. 

We can potentially think about making the id specification mandatory in the future, but that would be a backward incompatible breaking change and require active comms and grace period so we're not suddenly removing functionality that some user might be actually using.

For this PR, introducing the param as optional is only additive and therefore safe to do.

bc-tgomez: @bc-vincent-zhao @donald-nguyen-bc Hey ya'll 👋  Is the action here to set the param (`store_hash`) to optional? If so I tested that and it gives 7 errors. Not sure what changes to make if any here.

bc-vincent-zhao: @bc-tgomez the `store_hash` is still mandatory path.  

If you're referring to this particular line of change, it is to remove untrue statement that delete without any parameter deletes 1 page of data, because it deletes all coupons.

The real change of the PR is on line 241 where we add optional query parameter `id:in=1,2,3` so that users can use that to batch delete multiple coupons in one go by specifying all the ids of coupons to delete

WillemHoman (approved):

reference/marketing.v2.yml
bc-tgomez: Gotcha, makes perfect sense! Thanks as always @bc-vincent-zhao

bc-tgomez (approved):
Approving changes. Ready for merging.

PR #822, opened 2022-06-16
Devdocs 4144 update example code block

bc-traciporter:
# [DEVDOCS-4144](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4144)

## What changed?
I initially changed the redirects file and then noticed the other spec files had the same issue. So I updated the examples to relate more to that file.


slsriehl:
ticket dropped as duplicate

PR #824, opened 2022-06-16
Devdocs 4175 fork

markcmurphy:
# [DEVDOCS-](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-)

## What changed?
* thing_that_changed

### DIFF
--- a/reference/orders.v3.yml
+++ b/reference/orders.v3.yml
@@ -799,31 +799,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/MetaFieldCollectionResponse'
-              examples:
-                example-1:
-                  value:
-                    data:
-                      - permission_set: 'app_only'
-                        namespace: 'Sale Department'
-                        key: 'Staff Name'
-                        value: 'Ronaldo'
-                        description: 'Name of the sale staff.'
-                        resource_type: 'order'
-                        resource_id: '101'
-                        id: 1
-                        created_at: '2019-08-24T14:15:22Z'
-                        updated_at: '2019-08-24T14:15:22Z'
-                    meta:
-                      pagination:
-                        total: 1
-                        count: 1
-                        per_page: 50
-                        current_page: 1
-                        total_pages: 1
-                        links:
-                          previous: ''
-                          current: '?page=1&limit=50'
-                          next: ''
+              examples: {}
         '404':
           description: |
             The resource was not found.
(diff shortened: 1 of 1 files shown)


reference/orders.v3.yml
```diff
@@ -2784,79 +2718,22 @@ components:
...
-                previous: string
-                current: '?page=1&limit=50'
-                next: string
+      x-examples: {}
+      allOf:
+        - type: object
+          properties:
+            data:
+              $ref: '#/components/schemas/Metafield'
+        - $ref: '#/components/schemas/Meta'
```
markcmurphy, line 2731: Changed how the Meta model is referenced because it was nesting 'meta' within 'meta':

image

reference/orders.v3.yml
```diff
@@ -2909,22 +2786,78 @@ components:
...
-          type: string
+          type: integer
           description: |
             The unique identifier for the resource with which the metafield is associated.
-          maxLength: 10000000000
+          example: 0
+          readOnly: true
+      required:
+        - permission_set
+    MetafieldBase_Post:
```
markcmurphy, line 2803: Created a model for a metafield base that doesn't include `resource_id`. At least for orders, the `resource_id` seems like it is always going to be taken from the `order_id` path param to show the relationship:

image

bc-tgomez:
Hey @bigcommerce/orders team, could you all please review these changes and verify that they are accurate for Mark and myself?

reference/orders.v3.yml
```diff
@@ -2909,22 +2786,78 @@ components:
...
             - variant
+            - category
+          example: order
         resource_id:
-          type: string
+          type: integer
           description: |
             The unique identifier for the resource with which the metafield is associated.
-          maxLength: 10000000000
+          example: 0
```
bc-vincent-zhao, line 2799: 🍹 : may use `1` ? not sure if `0` is a valid resource id

bc-vincent-zhao (approved):
The changes in this PR looks sensible 👍 

Caveat: I'm not an expert on metafield, metafield should be horizontal feature across all domains, so people would probably expect the doc are same / very similar across metafields in catalog / orders/ customer / etc, so you can potentially compare this to the other existing docs and see how close they're.

reference/orders.v3.yml
```diff
@@ -2909,22 +2786,78 @@ components:
...
-            - category
             - order
             - brand
             - product
             - variant
+            - category
+          example: order
         resource_id:
-          type: string
+          type: integer
```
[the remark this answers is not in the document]
bc-tgomez: Good catch!

reference/orders.v3.yml
bc-tgomez: That works, thanks!

bc-tgomez (approved):
Approving --thanks for the revisions @markcmurphy and @bc-vincent-zhao  for the feedback! 🙇

PR #830, opened 2022-06-20
DEVDOCS-3755: Modified reference/orders.v3.yml Correct Refund Quote body example

bc-tgomez:
# [DEVDOCS-3755](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-3755)

## Correct Refund Quote body example
* Clean body example for Refund Quote in Orders V3


reference/orders.v3.yml
```diff
@@ -2268,6 +2268,14 @@ components:
         - order_id
         - items
       x-internal: false
+      x-examples:
+        example-1:
+          order_id: 1
```
slsriehl, line 2273:
```suggestion
          order_id: 1
```
use an order_id that is more like what they'll encounter.

traci707 (changes requested):
@bc-tgomez You are still missing the brackets.

{
  "order_id": 1,
  "items": 
[
  {
    "item_type": "SHIPPING",
    "item_id": 1,
    "amount": 1.99,
    "reason": "string"
  }
 ]
}

markcmurphy (approved):

PR #866, opened 2022-07-08
Devdocs 4296 updated keyword definition

traci707:
# [DEVDOCS-4296](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4296)

## What changed?
I tested the Get All Products endpoint using a brand name and then using a description for the keyword and did not receive a response.  But did receive a response when I used a name and an SKU. Therefore, I updated the definition to only include name and SKU.
See the screenshot of the slack conversation with Vlad, confirming my findings.


reference/catalog.v3.yml
```diff
@@ -304,8 +304,7 @@ paths:
             type: integer
         - name: keyword
           in: query
-          description: |
-            Filter items by keywords found in the `name`, `description`, or `sku` fields, or in the brand name.
+          description: 'Filter items by keywords found in the `name`, or `sku` fields'
```
markcmurphy, line 307:
```suggestion
          description: 'Filter items by keywords found in the `name` or `sku` fields'
```

markcmurphy (changes requested):

markcmurphy (approved):

PR #891, opened 2022-07-27
Fix error reported by heather in slack channel

traci707:
No ticket, see #dev-documentation slack channel
https://bigcommerce.slack.com/archives/C57SQ401K/p1658929913702279

## What changed?
Added the X-Auth-Token field to the Checkout API

PR #892, opened 2022-07-27
(no ticket): add x-auth-token to request runner...

slsriehl:
for add-a-consignment. Per Heather in Slack


slsriehl:
This is the same work as #891, repackaged for branch hygiene reasons. Next time I'll try to do better at preserving credit. Thanks everyone!

PR #895, opened 2022-08-01
DEVDOCS-4368: chore(customers): CUST-1745 add includes for segments

mjhmatt:
Adding docs for segments/shopper profiles

image
image
image


mjhmatt:
@bigcommerce/dev-docs These two fields that can be returned are in beta is there a particular format to display this?

PR #945, opened 2022-08-16
DEVDOCS-4007: [net new] Tax Rates Zones API

bc-andreadao:
# [DEVDOCS-4007](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4007)

This PR is related to:
-[PR 1613](https://github.com/bigcommerce/dev-docs/pull/1613): API Doc
-[PR 133](https://github.com/bigcommerce-labs/next-dev-center-prototype/pull/133): redirects


bc-andreadao:
@theromulans I had a quick question about a few things:
- Is it possible to delete all rates at once (or is the `id:in` query parameter required)? How about for deleting zones? 
- When creating a zone:
  - I assume the `name`, `price_display_settings`, and `locations` fields are the only required fields? 
  - Are there default values for `price_display_settings`? 
  - If a zone applies to all customers, does `customer_groups` return null?
- I put Info & Settings for the scope, not sure if correct

theromulans:
Great questions @bc-andreadao, thanks for reaching out.

> Is it possible to delete all rates at once (or is the id:in query parameter required)? How about for deleting zones?

No, we expect a list of IDs to be provided for all bulk delete requests. A request with an empty list is expected to always be rejected.

> I assume the name, price_display_settings, and locations fields are the only required fields?

During **create** the required fields will be `name` and `show_inclusive`.

Meanwhile for `locations`, yes, each tax zone will need at least one valid `location`.

Note: the default tax zone ("rest of the world" tax zone) does not have a location, but this is a special case and won't speak to anything a merchant can create at this stage (there's a feature request to add support for a "rest of the world" tax zone per customer group - but we can make any adjustments if/when this flow is supported).

> Are there default values for price_display_settings?

At the moment there is not, but in the interests of simplifying **create** requests we can have the following defaults:

`price_display_settings->show_both_on_detail_view` default: false
`price_display_settings->show_both_on_list_view` default: false

`show_inclusive` cannot have a default since neither option is ever more correct than the other.

I'll follow up on the implementation to make sure these are active (along with the existing default: true for `enabled`).

> If a zone applies to all customers, does customer_groups return null?

We would return an empty array for the `customer_groups` key. Does that sound reasonable?

> I put Info & Settings for the scope, not sure if correct

That's correct - all tax endpoints are currently under that scope, thanks for checking.

bc-andreadao:
@theromulans Thanks for the info! I made some edits to account for what you mentioned (required fields, empty array for customer groups, etc). Yes, it would be great if you could check on the default values. Hopefully the spec looks good from there

theromulans (approved):
Thanks @bc-andreadao

slsriehl:
@bc-andreadao get a peer review first, on the language etc. in theory we'd like to keep the final review very brief and make it mostly about the merge process.

reference/tax_rates_zones.v3.yml
```diff
@@ -0,0 +1,504 @@
...
+          items:
+            type: object
+            properties:
+              rate:
+                type: number
+                description: The tax rate that will be applied to the items in a tax class.
+                example: 5
+              tax_class_id:
+                type: integer
+                description: ID of a tax class. Each rate must be associated to a tax class. The rate will apply to all items associated with this tax class.
```
bc-traciporter, line 458:
```suggestion
                description: ID of a tax class. You must associate a tax rate with a tax class. The rate will apply to all the items related to this tax class.
```

reference/tax_rates_zones.v3.yml
```diff
@@ -0,0 +1,504 @@
...
+                example: 1
+    Tax_RatePOST:
+      type: object
+      properties:
+        tax_zone_id:
+          type: integer
+          description: ID of an associated tax zone. Tax rates must be associated to a tax zone.
```
bc-traciporter, line 465:
```suggestion
          description: ID of an associated tax zone. You must associate a tax rate with a tax zone.
```

reference/tax_rates_zones.v3.yml
```diff
@@ -0,0 +1,504 @@
...
+          example: 2
+        name:
+          type: string
+          description: 'The human-readable name for this tax zone. Displayed to merchant in the control panel and to shoppers, depending on store tax settings.'
```
bc-traciporter, line 469:
```suggestion
          description: 'The human-readable name for this tax zone. The name displays on the merchant's control panel and to shoppers, depending on store tax settings.'
```

bc-traciporter (commented):
I made editorial suggestions to remove passive voice.

bc-andreadao:
> At the moment there is not, but in the interests of simplifying **create** requests we can have the following defaults:
> 
> `price_display_settings->show_both_on_detail_view` default: false `price_display_settings->show_both_on_list_view` default: false
> 
> `show_inclusive` cannot have a default since neither option is ever more correct than the other.
> 
> I'll follow up on the implementation to make sure these are active (along with the existing default: true for `enabled`).

@theromulans I just wanted to do a quick follow up! For **creating** a tax zone, did we end up having defaults for `price_display_settings->show_both_on_detail_view` and `price_display_settings->show_both_on_list_view`?

theromulans:
@bc-andreadao Thanks for checking in. Yes, I can see that we have those default values in place and do not require those three fields to be present within the key set (though they can be).

That is, `enabled` and `price_display_settings->show_both_on_detail_view` and `price_display_settings->show_both_on_list_view`.

bc-andreadao:
> @bc-andreadao Thanks for checking in. Yes, I can see that we have those default values in place and do not require those three fields to be present within the key set (though they can be).

@theromulans Got it, thanks, I added the defaults values in

markcmurphy (dismissed):

slsriehl (approved):
most excellent. one link to the narrative docs removed so we can publish this first.

PR #983, opened 2022-09-20
Devdocs 4487 redirect urls via storefront cart

bc-traciporter:
# [DEVDOCS-4487]

## What changed?
Updated notes to include the Storefront Cart API

## Anything else?

[DEVDOCS-4487]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


valentindellangela (approved):
Thanks for updating the documentation

reference/carts.v3.yml
```diff
@@ -1005,7 +1005,7 @@ components:
               description: Given name for a gift certificate line item.
             theme:
               type: string
-              description: The theme of the gift certificate. The following options are available:`Birthday`, `Boy`, `Celebration`, `Christmas`, `General`, and `Girl`.
+              description: 'The theme of the gift certificate. The following options are available:`Birthday`, `Boy`, `Celebration`, `Christmas`, `General`, and `Girl`.'
```
bc-andreadao, line 1008: Are these the only possible values? If they are, we could document using `enum` (rather than putting them in the description).

bc-andreadao (commented):
just one suggestion about using `enum`

reference/carts.v3.yml
bc-traciporter: @bc-andreadao I think they are examples of some of the options but there can be more. Should I still use `enum`?

bc-andreadao: I did that for my Store Logs API, so my opinion is yes. Though maybe up to codeowners to decide

markcmurphy (approved):

PR #985, opened 2022-09-21
DEVDOCS-4449 Add Date Fields to Customs Information API

payaalaee:
[DEVDOCS-4449](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4449)

## What changed?

- Added documentation related to Customs Information `created_at` & `updated_at` date fields.


reference/shipping.v3.yml
```diff
@@ -410,6 +410,16 @@ components:
           example: true
         hs_codes:
           $ref: '#/components/schemas/harmonizedSystemCodes'
+        createdAt:
```
agchan12, line 413: While reading the [docs](https://github.com/bigcommerce/api-specs/pull/985/files#diff-8baa1a8874b11300860b8205908bd7256fd88d9ac522bff369b69380c7049f6aR128) it does appear that this will be part of the request body also which we don't want.

theromulans: Indeed. Does this mean we need to separate out request and response? Any thoughts on how best to proceed here @bc-andreadao?

reference/shipping.v3.yml
```diff
@@ -410,6 +410,16 @@ components:
...
+          type: string
+          description: Date and time of the customs information's creation. Read-Only.
```
theromulans, line 415: Is there a better approach to marking a field as read-only than just updating the description?

For example, I see in Tax Properties documentation we have `readOnly: true`. https://github.com/bigcommerce/api-specs/blob/9e603bdb4ca15b38b343acf0f86d719332e37f76/reference/tax_properties.v3.yml#L297

Would be worth updating the other existing item in this yml file too with this better approach, while we are here.

reference/shipping.v3.yml
```diff
@@ -410,6 +410,16 @@ components:
...
+          format: date-time
+          example: 2022-09-21T14:15:00Z
+        updatedAt:
+          type: string
+          description: Date and time when the customs information was last updated. Read-Only.
+          format: date-time
+          example: 2022-09-21T14:15:00Z
```
theromulans, line 422: Interesting - I wonder if the Tax Properties docs are incorrect here. Any thoughts @bc-andreadao ?

`2022-07-21T19:33:57+00:00` (Tax Properties)
`2018-05-07T20:14:17Z` (Tax Classes)
`2022-09-21T14:15:00Z` (Here)

reference/shipping.v3.yml
```diff
@@ -382,8 +382,8 @@ components:
       title: customsInformation
       description: Data about the customs information object.
       type: object
-      minProperties: 5
-      maxProperties: 5
+      minProperties: 7
+      maxProperties: 7
```
theromulans, line 386: I'm not familiar with this - what does this describe? That all fields are required?

reference/shipping.v3.yml
bc-andreadao: @theromulans I'm actually getting `2022-09-23T21:50:38+00:00` format for all 3 APIs, so I changed it here and created a new ticket for tax classes. Thanks for catching that!

reference/shipping.v3.yml
bc-andreadao: I see what you mean. I just separated them out by creating a`customsInformation_request`.

reference/shipping.v3.yml
bc-andreadao: Yeah, I think it was meant to describe that the fields are required (not sure why it was documented like this before). I marked the fields as required in `customsInformation_request` .

reference/shipping.v3.yml
bc-andreadao: Yes, the readOnly flag looks better, thanks for mentioning. Just added them to all the read-only fields.

bc-andreadao (approved):

bc-andreadao:
@payaalaee  Thanks for the help! Let me know if there are other changes or new fields before GA date

reference/shipping.v3.yml
```diff
@@ -319,9 +319,10 @@ components:
         - type: object
           properties:
             id:
-              description: Shipping-method ID. READ-ONLY
+              description: Shipping-method ID.
```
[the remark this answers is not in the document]
bc-andreadao: @agchan12 Thanks for catching that, they are removed now

agchan12 (approved):

benpratt77 (approved):

theromulans (approved):
A great step forward, thanks for your help and thoughts @bc-andreadao!

bc-andreadao:
Changes will appear live on the GA date 🥳

reference/shipping.v3.yml
theromulans: Just wanted to point out that `Z` and +`00:00` are essentially the same thing, but good to reflect what we're producing from an API standpoint.

bc-andreadao: Sounds good. If you need a PR to follow, I made changes [here](https://github.com/bigcommerce/api-specs/pull/1015)

theromulans:
Hi @bc-andreadao @payaalaee, I just noticed that these fields are appearing as `createdAt` and `updatedAt` rather than `created_at` and `updated_at`.

Screen Shot 2022-10-21 at 11 05 59 am

I've checked our code and it seems the underscore approach is expected - is that accurate?

Any chance we can fix this up? Cheers.

bc-andreadao:
@theromulans Ah yes, it should be underscored (thanks for that, I must have missed it). I created a PR [here](https://github.com/bigcommerce/api-specs/pull/1023) to fix it up, should be fixed shortly.

PR #996, opened 2022-09-28
DEVDOCS-4209: [maintenance] product video

bc-andreadao:
# [DEVDOCS-4209]

Fixed description. Should work for the purposes of this ticket, since the request and response for POST /catalog/products are correct. 

Other methods for the endpoint aren't accurate at the moment, may change for a future bug fix (see slack conversation [here](https://bigcommerce.slack.com/archives/C9EUJ8LCC/p1664398230192769))


reference/catalog.v3.yml
```diff
@@ -903,7 +903,8 @@ paths:
         - 250 characters product name length.
 
         **Usage Notes**
-        * `POST` requests to `/products` accepts a single `video` object; to send an array of video objects, see: `/products/{product_id}/videos`.
+        * `POST` requests to `/products` accept a `video` array. 
+        * `POST` requests to `/products/{product_id}/videos` accept a video object.
```
markcmurphy, line 907:
```suggestion
        * `POST` requests to `/products/{product_id}/videos` accept a `video` object.
```

bc-tgomez: My only suggestion for the usage notes would be to make it more conversational like:
Send a Post request at `/...` to accept a `video` object. Otherwise, no issues here! 😺

bc-andreadao: good catch 🎈

markcmurphy (approved):

PR #1013, opened 2022-10-12
ORDERS-4923 Gift certificate consignment POST

WillemHoman:
# [DEVDOCS-4552](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4552)

## What changed?
Support Gift Certificate consignments in POST /v2/orders

## Anything else?
Engineering ticket: [ORDERS-4914](https://bigcommercecloud.atlassian.net/browse/ORDERS-4914)

@bigcommerce/dev-docs @bigcommerce/orders

### DIFF
--- a/reference/orders.v2.oas2.yml
+++ b/reference/orders.v2.oas2.yml
@@ -485,6 +485,25 @@ paths:
                     country: United States
                     country_iso2: US
                     email: janedoe@example.com
+              Gift Certificate Consignment:
+                value:
+                  customer_id: 11
+                  consignments:
+                    email:
+                      gift_certificates:
+                        - recipient_email: jane@example.com
+                          line_items:
+                            gift_certificate_id: 184
+                  billing_address:
+                    first_name: Jane
+                    last_name: Doe
+                    street_1: 123 Main Street
+                    city: Austin
+                    state: Texas
+                    zip: '78751'
+                    country: United States
+                    country_iso2: US
+                    email: janedoe@example.com
         required: true
       responses:
         '200':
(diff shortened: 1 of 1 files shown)


reference/orders.v2.oas2.yml
```diff
@@ -485,6 +485,25 @@ paths:
...
                     email: janedoe@example.com
+              Gift Certificate Consignment:
+                value:
+                  customer_id: 11
+                  consignments:
+                    email:
+                      gift_certificates:
+                        - recipient_email: jane@example.com
+                          line_items:
+                            gift_certificate_id: 184
```
danielquachnz, line 496: I must've missed this before. This is a different format to our GET isn't it? IIRC, we had the URL to the order product and then to get the GC id you needed to seperately query the order product to get the GC id column

reference/orders.v2.oas2.yml
```diff
@@ -5058,6 +5077,31 @@ components:
...
+        line_items:
+          type: array
+          minItems: 1
+          items:
+            type: object
+            properties:
+              gift_certificate_id:
+                description: ''
+                example: 2
+                type: integer
```
donald-nguyen-bc, line 5101: maybe this field should be marked as "required" as well

donald-nguyen-bc (approved):
👍

reference/orders.v2.oas2.yml
WillemHoman: I think the current format for `GET` is correct as the `POST` creates an `order_product` which references the gc, in the same way creating an order with a catalog product id will create an `order_product` which references the product

reference/orders.v2.oas2.yml
WillemHoman: https://github.com/bigcommerce/api-specs/pull/1013/commits/583247189e29a248b7e000b54902008ae16e584c

slsriehl (approved):
Merging this into DEVDOCS feature branch.

PR #1029, opened 2022-10-28
Devdocs 4553 update timestamp format

bc-traciporter:
# [DEVDOCS-4553]

## What changed?
updated timestamp format

## Anything else?

[DEVDOCS-4553]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/catalog.v3.yml
```diff
@@ -21292,6 +21361,38 @@ components:
...
+          id: 0
+          product_id: 0
+          sku: string
+          sku_id: 0
+          option_values:
+            - option_display_name: Color
+              label: Beige
+              id: 146
+              option_id: 151
+          calculated_price: -1.7976931348623157e+308
```
bc-andreadao, line 21394: Was the calculated price intentional? We could make it 0

bc-andreadao:
For the Sites API, I tested the [GET Sites](https://developer.bigcommerce.com/api-reference/4981c78979bd3-get-sites) endpoint, and got that the date format is actually `2021-12-15T05:05:48Z`. I assume that all the endpoints in the Site API are in that format, so maybe reach back out to the engineering team to see if this was intended 

response

reference/catalog.v3.yml
markcmurphy: I updated the value, so I'm resolving the convo

markcmurphy (approved):

PR #1056, opened 2022-11-17
DEVDOCS-4605: [update] Payments, add bigpay_token

bc-traciporter:
# [DEVDOCS-4605]

## What changed?
Added bigpay_token

## Anything else?

[DEVDOCS-4605]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/orders.v3.yml
```diff
@@ -2043,9 +2043,51 @@ components:
...
                   example: 35.42
-      title: Transaction
-      x-internal: false
       type: object
+      title: ''
+      x-examples:
+        Example 1:
+          event: purchase
+          method: credit_card
+          amount: -3.402823669209385e+38
```
bc-tgomez, line 2052: Thinking that this here needs to be adjusted.

bc-tgomez (commented):
Just want to take a look at that one comment but other than that, I think we're good to go. 😄

reference/orders.v3.yml
bc-traciporter: Good find @bc-tgomez. I made the number positive.

slsriehl (changes requested):
Let's wait on @bc-keith to make sure we're synced up with Payments. Presumably there will also be changes to at least a few of payment_processing, payment_methods, and customers.v3 list stored instruments.

slsriehl (approved):
Looks good. Nice additional changes.

PR #1062, opened 2022-11-23
SHIPPING-1607 altering the documentation after adding in the channel id

gazzayeatman:
# DEVDOCS-4570

## What changed?
- Removes cart_id as a field as it is no longer there
- Adds some description to the request_context field to describe that uses the request context

## Anything else?
Related branch
https://bigcommercecloud.atlassian.net/browse/SHIPPING-2535

@bigcommerce/shipping

### DIFF
--- a/reference/shipping_provider.yml
+++ b/reference/shipping_provider.yml
@@ -1984,14 +1984,12 @@ components:
                     type: string
             customer_id:
               type: integer
-        cart_id:
-          type: string
         store_id:
           type: string
         request_context:
           type: object
           title: Request Context
-          description: A collection of Reference Value objects.
+          description: A collection of Reference Value objects. Two examples of this are the cart_id and the channel_id.
           properties:
             reference_values:
               type: array
--- a/reference/shipping_provider.yml
+++ b/reference/shipping_provider.yml
@@ -1989,7 +1989,7 @@ components:
         request_context:
           type: object
           title: Request Context
-          description: A collection of Reference Value objects. Two examples of this are the cart_id and the channel_id.
+          description: A collection of Reference Value objects.
           properties:
             reference_values:
               type: array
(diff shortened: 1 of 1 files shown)


slsriehl:
@gazzayeatman please tag in someone from Shipping to approve these specs from a technical POV. Then @bc-andreadao will merge to our ticket branch so we can work on the language and narrative docs. Thanks y'all!

reference/shipping_provider.yml
```diff
@@ -1984,14 +1984,12 @@ components:
...
               type: integer
-        cart_id:
-          type: string
         store_id:
           type: string
         request_context:
           type: object
           title: Request Context
-          description: A collection of Reference Value objects.
+          description: A collection of Reference Value objects. Two examples of this is the cart_id and the channel_id.
```
bc-andreadao, line 1992: @gazzayeatman One question I had was whether this would be a valid example for `request_context`: 

`"request_context": {
   "reference_values": [
      {
         "name": "channel_id",
          "value": "1"
      },
      {
         "name": "cart_id",
          "value": "abc"
      },
   ]
}     
`

gazzayeatman: Correct, it would. The default channel ID is also 1

bc-luke (approved):
LGTM

agchan12 (approved):

benpratt77 (approved):

reference/shipping_provider.yml
theromulans, line 1992: 🍻 Would we want to mark these terms in any special way @bc-andreadao ? Unsure if we have a common approach here.

For example `cart_id` and `channel_id` would be how I would do it in this GitHub comment 😆

theromulans, line 1992: > Two examples of this is the cart_id and the channel_id.

...does this read better...?

> Two examples of this are the cart_id and the channel_id.

bc-andreadao: @theromulans I changed it so that the examples show up in the schema: (looks better in my opinion) 

examples

theromulans: That's awesome, thanks @bc-andreadao.

theromulans (approved):

PR #1095, opened 2022-12-16
DEVDOCS-4673: [update] Catalog API, changed max value for inventory

bc-andreadao:
# [DEVDOCS-4673]

Changed max value for `inventory_level` and `inventory_warning_level` 

[DEVDOCS-4673]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EgorAleshin (approved):

bc-tgomez (approved):

reference/catalog.v3.yml
```diff
@@ -23904,13 +23904,13 @@ components:
           description: |
             A product can be added to an existing brand during a product /PUT or /POST.
         inventory_level:
-          maximum: 1000000000
+          maximum: 2147483647
```
slsriehl, line 23907: @bc-andreadao are these values in bytes? String lengths? Do we specify the unit anywhere?

EgorAleshin: @slsriehl this is INT_MAX value https://learn.microsoft.com/en-us/cpp/cpp/integer-limits?view=msvc-170

bc-andreadao: @slsriehl I believe the unit is implied because the type is integer. Or else we would have to move it to the description

stanzikratelbc: @bc-andreadao I verified the maximum value from:  https://learn.microsoft.com/en-us/cpp/cpp/integer-limits?view=msvc-170, this looks good

stanzikratelbc (approved):
LGTM. Ready to merge

PR #1101, opened 2022-12-21
DEVDOCS-4575: [update] Catalog V3 API, update request response schemas

bc-andreadao:
# [DEVDOCS-4575]

- Removed `id` from POST request bodies. Sometimes, doing this also removed `id` from responses (that need `id`), so I had to add `id` in other places of the spec. 

- Metafield endpoints:
   1.  Removed unneeded fields from POST and PUT request bodies
   2.  Reordered fields for response examples 
   3.  Fixed `resource_type` for response examples (should correspond to the endpoint) 

[DEVDOCS-4575]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1108, opened 2022-12-29
DEVDOCS-4684: [update] Catalog, revise bulk pricing rule amount definition

bc-traciporter:
# [DEVDOCS-4684]

## What changed?
changed the definition of amount
See dev docs ticket for information

## Anything else?

[DEVDOCS-4684]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/catalog.v3.yml
```diff
@@ -22856,9 +22856,9 @@ components:
           x-required:
             - post
         amount:
-          type: integer
+          type: number
           description: |-
-            The discount can be a fixed dollar amount or a percentage. For a fixed dollar amount enter it as an integer and the response will return as an integer. For percentage enter the amount as the percentage divided by 100 using string format. For example 10% percent would be “.10”. The response will return as an integer.
+            The discount can be a fixed dollar amount or a percentage. For a fixed dollar amount enter it as an number and the response will return as an number. For percentage enter the amount as the percentage divided by 100 using string format. For example 10% percent would be “.10”. The response will return in decimal format.
```
bc-tgomez, line 22861:
```suggestion
            The discount can be a fixed dollar amount or a percentage. For a fixed dollar amount enter it as a number and the response will return as a number. For percentage, enter the amount as the percentage divided by 100 using string format. The response will return in decimal format. For example, 10% percent would be “.10”.
```

slsriehl:
```suggestion
            You can express the discount as either a fixed dollar amount or a percentage. To use a fixed dollar amount, send a number; the response will return a number. To use a percentage, divide the percentage by 100 and send the result in string format. For example, represent 10% as “.10”. The response will return a float.
```

reference/catalog.v3.yml
```diff
@@ -22856,9 +22856,9 @@ components:
           x-required:
             - post
         amount:
-          type: integer
+          type: number
```
slsriehl, line 22859: this should be number or string.

slsriehl (changes requested):
1. This entry occurs twice in the spec file, and this schema is used in several larger schemas. Because the datatypes, required fields, etc, are quite different depending on whether the bulk pricing rule is part of a request or response, I recommend paring this down and defining the request and the response separately using allOf. If we don't represent this complexity, we'll keep getting tickets.
2. The ticket has another ask, which is to revise the `quantity_min` property.

Thanks for the work, everyone.

bc-traciporter:
@bc-andreadao @slsriehl  recommends having different schemas for requests and responses for Product Bulk Pricing Rules endpoints and the other Catalog endpoints as well. I know you have a PR that changes the structure for many of these endpoints. I would like to wait until your PR goes through first before sending this PR though.

bc-andreadao:
> @bc-andreadao @slsriehl recommends having different schemas for requests and responses for Product Bulk Pricing Rules endpoints and the other Catalog endpoints as well. I know you have a PR that changes the structure for many of these endpoints. I would like to wait until your PR goes through first before sending this PR though.

@bc-traciporter Yeah my [PR 1101](https://github.com/bigcommerce/api-specs/pull/1101) changes the catalog api endpoints so that request and responses have the correct schema. Just bumped it for final review

slsriehl:
FYI @bc-traciporter 1101 has been merged to staging.

reference/catalog.v3.yml
```diff
@@ -11953,7 +11953,7 @@ paths:
                         minimum: 0
                         type: integer
                         description: |
-                          The minimum inclusive quantity of a product to satisfy this rule. Must be greater than or equal to zero.
+                          The minimum inclusive quantity of a product to satisfy this rule. Must be greater than or equal to zero. For `fixed` rules, the minimum quantity can't be less than two.
```
slsriehl, line 11956: replace this single straight quote with an apostrophe character.

reference/catalog.v3.yml
```diff
@@ -21930,13 +21874,56 @@ components:
...
         amount:
-          type: integer
+          oneOf:
+            - type: number
+              example: 10
+            - type: string
+              example: '.10'
           description: |-
-            The discount can be a fixed dollar amount or a percentage. For a fixed dollar amount enter it as an integer and the response will return as an integer. For percentage enter the amount as the percentage divided by 100 using string format. For example 10% percent would be “.10”. The response will return as an integer.
+            You can express the discount as either a fixed dollar amount or a percentage. To use a fixed dollar amount, send a number; the response will return a number. To use a percentage, divide the percentage by 100 and send the result in string format. For example, represent 10% as “.10”. The response will return a float.
```
slsriehl, line 21883: we have 3 types of discount amount, and it's not clear to me whether the fixed dollar amount expresses price discounts or fixed discounts.

PR #1115, opened 2023-01-04
DEVDOCS-4668: [maintenance] revoke graphql tokens

bc-andreadao:
# [DEVDOCS-4668]

Added note about revoking only compromised tokens


rafa-avila-bc (dismissed):

bookernath (dismissed):
LGTM, made a suggestion for improved verbiage.

reference/storefront_tokens.v3.yml
```diff
@@ -173,7 +173,7 @@ paths:
       tags:
         - API Token
       summary: Revoke a Token
-      description: Revoke access for a Storefront API token. Only revoke compromised tokens under emergency situations. Uncompromised short-lived tokens do not need to be revoked and should be allowed to naturally expire.
+      description: Revoke access for a Storefront API token. Only revoke compromised tokens under emergency situations. Allow uncompromised short-lived tokens to naturally expire, as these do not need to be revoked.
```
bc-traciporter, line 176: You can take or leave the suggestion below.

Revoke access for a Storefront API token. Only revoke compromised tokens under emergency situations. Let uncompromised short-lived tokens naturally expire, as you do not need to revoke these.

bc-tgomez: Agree with @bc-traciporter on the rewording suggestion, but other than that, looks splendid!

bc-tgomez (dismissed):

PR #1146, opened 2023-02-16
feat(tax): DEVDOCS-4783 Introduce should subtract store tax

bc-msung:
# [DEVDOCS-4783](https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4783)

## What changed?
Update the `tax_settings` documentation  with a new field `should_subtract_store_tax`
This will allow the merchants to both read and set the new `should_subtract_store_tax` via the bigcommerce public API

## Anything else?
PR: https://github.com/bigcommerce/bigcommerce/pull/52265
JIRA Ticket: https://bigcommercecloud.atlassian.net/browse/TAX-1421
JIRA Ticket: https://github.com/bigcommerce/bigcommerce/pull/52265
(naming change from `tax_should_unwind_store_tax` -> `should_subtract_store_tax`)

ping @bigcommerce/shipping @bc-andreadao @bigcommerce/devdocs-codeowners 

[DEVDOCS-4783]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/reference/tax_settings.v3.yml
+++ b/reference/tax_settings.v3.yml
@@ -143,6 +143,10 @@ components:
             - FIXED
             - BASIC
             - DISABLE
+        should_subtract_store_tax:
+          default: true
+          type: boolean
+          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the item's store tax rate before calculating tax using the shopper's tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopper's location. The tax-exclusive amount may vary among tax zones.    
     Tax_Settings_Req:
       type: object
       properties:
--- a/reference/tax_settings.v3.yml
+++ b/reference/tax_settings.v3.yml
@@ -146,7 +146,7 @@ components:
         should_subtract_store_tax:
           default: true
           type: boolean
-          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the item's store tax rate before calculating tax using the shopper's tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopper's location. The tax-exclusive amount may vary among tax zones.    
+          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the item's store tax rate before calculating tax using the shopper's tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopper's location. The tax-exclusive amount may vary among tax zones. These calculations are relevant for tax pricing and tax quotations that use basic tax.    
     Tax_Settings_Req:
       type: object
       properties:
@@ -177,7 +177,7 @@ components:
         should_subtract_store_tax:
           default: true
           type: boolean
-          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the item's store tax rate before calculating tax using the shopper's tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopper's location. The tax-exclusive amount may vary among tax zones.
+          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the item's store tax rate before calculating tax using the shopper's tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopper's location. The tax-exclusive amount may vary among tax zones. These calculations are relevant for tax pricing and tax quotations that use basic tax.
     MetaOpen:
       title: Response meta
       type: object
--- a/reference/tax_settings.v3.yml
+++ b/reference/tax_settings.v3.yml
@@ -31,7 +31,7 @@ paths:
         '200':
           description: OK
           content:
-            application/json: 
+            application/json:
               schema:
                 type: object
                 properties:
@@ -170,6 +170,10 @@ components:
             - FIXED
             - BASIC
             - DISABLE
+        should_subtract_store_tax:
+          default: true
+          type: boolean
+          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the item's store tax rate before calculating tax using the shopper's tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones, only the tax amount will vary based on the shopper's location. The tax-exclusive amount may vary between tax zones.
     MetaOpen:
       title: Response meta
       type: object
--- a/reference/tax_settings.v3.yml
+++ b/reference/tax_settings.v3.yml
@@ -173,7 +173,7 @@ components:
         should_subtract_store_tax:
           default: true
           type: boolean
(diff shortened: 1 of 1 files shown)


reference/tax_settings.v3.yml
```diff
@@ -141,6 +141,9 @@ components:
             - FIXED
             - BASIC
             - DISABLE
+        tax_should_unwind_store_tax:
+          type: boolean
+          description: Determines tax unwinding behaviour where when active, it takes away the store tax rate and applies the shopper tax rate.
```
benpratt77, line 146: The experiment description is as follows:
`When enabled, we avoid rounding during tax unwinding processes`
I would use something similar to this.

bc-msung: updated 👍

reference/tax_settings.v3.yml
```diff
@@ -141,6 +141,9 @@ components:
             - FIXED
             - BASIC
             - DISABLE
+        tax_should_unwind_store_tax:
+          type: boolean
+          description: When enabled, we avoid rounding during tax unwinding processes
```
benpratt77, line 146: can we add a full stop :P

bc-msung: updated 👍

bc-andreadao:
@bc-msung When you're ready, let me know the fixes/changes from [TAX-1702](https://bigcommercecloud.atlassian.net/browse/TAX-1702)

reference/tax_settings.v3.yml
```diff
@@ -170,6 +170,9 @@ components:
             - FIXED
             - BASIC
             - DISABLE
+        should_subtract_store_tax:
+          type: boolean
+          description: When enabled, we avoid rounding during subtract store tax processes.
```
bc-andreadao, line 175:
```suggestion
          description: If a merchant enters **tax-inclusive** prices, this setting indicates whether a store will update the **product's list price** on the storefront to reflect the shopper's tax zone. When enabled, the store subtracts the tax rate  before calculating tax using the shopper's tax zone. When disabled, the tax-inclusive price is the same across all tax zones, and the store updates the **product base price** to reflect differences in tax rates across zones.
```

bc-andreadao:
@bc-msung Thanks for the update. Could you also provide a default value? Is the default `false` in both cases (when a merchant enters prices inclusive OR exclusive of tax)?

reference/tax_settings.v3.yml
bc-andreadao: Need someone to check the accuracy of my statement

slsriehl: i'm not sure how `and the store updates the **product base price** to reflect differences in tax rates across zones` works with tax-inclusive pricing. Surely it's not an update to the catalog product object.

theromulans: Thanks @bc-andreadao. Certainly the original description isn't accurate - any rounding concerns are a separate matter from this feature, which is simply whether or not we subtract store tax.

I'm not sure I would speak to specific implementation or describe it the way above. How about some minor adjustments:

> This setting applies only if a merchant enters **tax-inclusive** prices.

👍 I think it's important to point this out - very nice.  

It may also be important to point out that it's only relevant to basic tax - but this may introduce further descriptive complexity since basic tax is always used for **tax pricing** but can optionally be used for **tax quotation**.

> When enabled, the store subtracts the item's store tax rate before calculating tax using the shopper's tax zone. 

I've adjusted the wording here slightly. A few notes:
* **Item** instead of using **Product** given this approach applies to non-product items too (e.g. shipping, wrapping).
* **Store tax rate** to hopefully more descriptively answer "Which tax rate?" (instead of just "the tax rate").

Alternatively:

> When enabled, the store subtracts the item's store tax rate before calculating tax using the shopper's tax zone. The tax-exclusive amount will be the same across all tax zones.

This version describes the tax exclusive amount being consistent, to hopefully better compare the two options.

> When disabled, the tax-inclusive price remains the same across all tax zones, only the tax amount will vary based on the shopper's location. The tax-exclusive amount may vary between tax zones.

So... perhaps...

> This setting applies only if a merchant enters **tax-inclusive** prices. When enabled, the store subtracts the item's store tax rate before calculating tax using the shopper's tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones, only the tax amount will vary based on the shopper's location. The tax-exclusive amount may vary between tax zones.

reference/tax_settings.v3.yml
```diff
@@ -170,6 +170,9 @@ components:
             - FIXED
             - BASIC
             - DISABLE
+        should_subtract_store_tax:
```
theromulans, line 173: 😅 Just to make doubly sure, since I know there has been confusion here recently, the default is `true` (that is, this is the current state for all stores).

reference/tax_settings.v3.yml
bc-andreadao: Thanks @theromulans, I found that super helpful ! 

> It may also be important to point out that it's only relevant to basic tax - but this may introduce further descriptive complexity since basic tax is always used for **tax pricing** but can optionally be used for **tax quotation**.
 
Would it be accurate (and simpler) if I flipped this and said that the setting "doesn't apply if a merchant uses a third-party tax provider for tax quotations"? I was looking at this [slack conversation](https://bigcommerce.slack.com/archives/C44T5JWP2/p1679311143545369?thread_ts=1674643606.762549&cid=C44T5JWP2), but wasn't sure what the final conclusion was.  

Could you explain what you mean when you say that there would be descriptive complexity if I mentioned basic tax?

Also, can the `should_subtract_store_tax` setting affect what shoppers see on a storefront **before checkout**? 
For example, if the following were true:
- a merchant enters tax-inclusive prices,
- the storefront displays prices inclusive of tax,
- the tax zone of the shopper's customer group is different than that of the merchant
then do the prices displayed on the product pages depend on the `should_subtract_store_tax` setting?

reference/tax_settings.v3.yml
bc-andreadao: Got it, thanks, added the default value!

bc-msung:
updated the description and added in default values as well.

reference/tax_settings.v3.yml
theromulans: @bc-andreadao Awesome, thanks for your insightful questions as always!

**Tax pricing** happens outside the context of a cart, for individual prices shown on the storefront - these prices are considered estimates or marketing prices. **Tax quotation** is when we have an entire cart context and we are figuring out how much tax the shopper will pay.

All stores always use basic tax for tax pricing... but not all stores use basic tax for tax quotation. This setting is relevant in both cases: tax pricing and tax quotation.

Today the setting can affect the storefront experience, tax pricing. Specifically, when a customer logs in we may assign a different tax zone to them, and hence store tax will be subtracted before the customer group tax zone's rates are added.

In the near future we have our _tax zone price display for storefront_ project that will also utilise GeoIP to determine a better match for an anonymous shopper on the storefront, meaning this setting will have higher relevance for tax pricing.

So the complexity I was speaking to is mainly around getting into the weeds of this behaviour. It's also quite dependant on how the merchant has configured their tax zones. I would safely guess that the majority of stores today do not have tax zones configured in a way that really makes a difference here.

bc-andreadao: Thanks @theromulans ! Just added a sentence about basic tax.

theromulans: Loving it, thanks @bc-andreadao!

PR #1150, opened 2023-02-17
(no ticket): [tools] MDX linter, create + add github action

slsriehl:
# (no ticket): [tools] MDX linter, create + add github action

## What changed?
* add npm script 
* add eslint config
* add GitHub action

## Anything else?
cc @matthewvolk


slsriehl:
I think this should fail because there are no mdx files present

PR #1153, opened 2023-02-20
(no ticket): [migrate] Remove path params

slsriehl:
# (no ticket): [migrate] Remove path params

## What changed?
* this is for the record and to squash

## Anything else?


slsriehl:
doing this differently

PR #1179, opened 2023-03-02
PROMO-702: [update] Channels, tighten definitions

leeBigCommerce:
# PROMO-702

## What changed?
* `data` and `meta` fields within `multiple_channels_without_currencies_resp` made required
* `id` and `name` fields within `ChannelWithoutCurrencies` made required

## Anything else?

I found some mistakes here while moving to use auto-generated Type definitions for some promotions view work.
The above values were portrayed as potentially `undefined` (or nullable), which I don't believe is a true representation of the API.

There are likely to a lot more values that are actually required, but that would need someone more knowledgeable of the Channels domain to look into.


slsriehl:
Hi @leeBigCommerce. Some changes to the spec files are coming in the next few days. I've marked his PR as draft.

In the meantime, if PROMO-702 doesn't have an associated CSS project ticket, please ping #operational-readiness so they can get that and a child DEVDOCS ticket made.

slsriehl (approved):
Nice step forward, @leeBigCommerce.

PR #1199, opened 2023-03-06
DEVDOCS-4719: [migrate] 

bc-andreadao:
# [DEVDOCS-4719]

APIs in this PR: 
- Storefront Carts
- Carts V3 API 

Move `x-examples` so that they render  (`x-examples` don't render). 

Note that I didn't move the request body x-examples for the Cart V3 API's Create a Cart endpoint. These `x-examples` seem to almost duplicate the `examples` that do render. I didn't delete them because we should test the `x-examples` and the `examples` before seeing which to delete.


bc-andreadao:
@slsriehl  Request body examples that don't show on api-tester
- Update Customer ID 
- Update Channel Cart Settings
- Update Global Cart Settings

slsriehl (approved):
attempt several fixes

PR #1210, opened 2023-03-10
Fix the case inconsistency issue for the DetailedErrors model

kzhang-dsg:
Fix the case inconsistency issue for the DetailedErrors model in catalog.v3. The inconsistent case causes problems in code generators.

# [DEVDOCS-] No ticket

## What changed?
* Changed the `detailedErrors` to `DetailedErrors` to match the cases in other API spec files. 

## Anything else?
No


slsriehl (approved):
LGTM

slsriehl:
See the following PR for next merge into publication stream:
* #1275

PR #1233, opened 2023-03-27
(no ticket): [external] Webhooks, fix header field to indicate support for custom key-value pairs

kzhang-dsg:
The API document has hard coded
```
{
    custom: "string"
}
```
as the headers, which is not correct. the headers should be a free formed key value pair.


slsriehl:
@kzhang-dsg can you rebase this one?

slsriehl (approved):
Thanks, @kzhang-dsg!

PR #1249, opened 2023-04-04
Type product_Put_Collection does not contain IDs to update 

kzhang-dsg:
A batch update of catalog products must contain identifiers


slsriehl:
For catalog change, see the following:
* #1275

PR #1251, opened 2023-04-04
DEVDOCS-4846: [update] OrdersV3,  include note on payment_actions/refunds and /refund_quotes

bc-tgomez:
Add note on how order refunds are processed


bc-andreadao (approved):

reference/orders.v3.yml
```diff
@@ -161,6 +161,10 @@ paths:
         Requires at least one of the following scopes:
         * `store_v2_orders`
         * `store_v2_transactions`
+
+        **Note:**
+        Order refunds are processed consecutively. Processing synchronous refunds on an order is not yet supported.
```
bc-traciporter, line 166:
```suggestion
        Order refunds are processed consecutively. Processing synchronous refunds on an order are not yet supported.
```

reference/orders.v3.yml
```diff
@@ -223,6 +227,10 @@ paths:
...
```
bc-traciporter, line 232:
```suggestion
        Order refunds are processed consecutively. Processing synchronous refunds on an order are not yet supported.
```

bc-traciporter (changes requested):
grammatical changes suggested

PR #1286, opened 2023-04-26
removed address and added ShippingAddress for consignments

bc-traciporter:
Changed "address" to "ShippingAddress" for adding and updating Consignments


valentindellangela:
Hi @bc-traciporter, is there any reason you're updating this field?
Following Bopis we have created a new field called `address` so it's more generic and can work for both Shipping and pickup consignment so I would suggest to keep it as is.

animesh1987:
@valentindellangela this was a question raised in slack from memory, looking at it now seems like no one got to it. I will tag you in the slack thread for visibility.

bc-traciporter:
@animesh1987 @valentindellangela I will close this PR. I tested this endpoint again; the only issue is that the request needs to be enclosed in an array.  Once I added the brackets, it worked. I will create a new PR for this change.

PR #1289, opened 2023-04-27
DEVDOCS-4459: [update] Customer Groups, Get and Delete endpoints need 400 response 

bc-tgomez:
Add a 400 response to both endpoints:
* GET /v2/customer_groups/{group_ID}
* DELETE /v2/customer_groups/{group_ID}


bc-traciporter (approved):
LGTM

PR #1297, opened 2023-05-02
Rename `url` to `custom_url` in Category Model

kzhang-dsg:
According to the APi response, the `url` property should be `custom_url` for Category Model


slsriehl:
@kzhang-dsg we've split the catalog file up into several parts and are on the cusp of building the site using those files. You are welcome to resubmit these changes to one or more of the files in `reference/catalog/*`.

PR #1299, opened 2023-05-02
Rename `url` to `custom_url` in Category Model

kzhang-dsg:
According to the APi response, the `url` property should be `custom_url` for Category Model

### DIFF
--- a/reference/catalog.v3.yml
+++ b/reference/catalog.v3.yml
@@ -21587,7 +21587,7 @@ components:
         - $ref: '#/components/schemas/default_product_sort'
         - type: object
           properties:
-            url:
+            custom_url:
               $ref: '#/components/schemas/Url'
       x-examples: {}
     CategoryUuidData:
@@ -21670,7 +21670,7 @@ components:
               type: boolean
             image_url:
               type: string
-            url:
+            custom_url:
               $ref: '#/components/schemas/Url'
     CategoryDataPUT:
       allOf:
@@ -21692,7 +21692,7 @@ components:
     Url:
       type: object
       properties:
-        path:
+        url:
           type: string
         is_customized:
           type: boolean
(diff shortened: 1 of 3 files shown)


bc-traciporter:
@kzhang-dsg Thank you for submitting this PR.  The line you indicated refers to a model that is not called in the specification file. Changing this line would not change what is displayed in our developer docs.

bc-traciporter:
@kzhang-dsg After looking at it again, you are correct.  We will merge your changes.

bc-traciporter (approved):
LGTM

slsriehl (changes requested):
@kzhang-dsg Until we dedupe the new Catalog files, we have repeated schemas. Please make these changes in the `category-trees_catalog.v3.yml` file as well.

bc-traciporter:
@slsriehl I updated the `categories-trees_catalog.v3.yml file. Can you review it?

reference/catalog.v3.yml
```diff
@@ -21587,7 +21587,7 @@ components:
         - $ref: '#/components/schemas/default_product_sort'
         - type: object
           properties:
-            url:
+            custom_url:
```
slsriehl, line 21590:
```suggestion
            url:
```

revert

reference/catalog.v3.yml
```diff
@@ -21692,7 +21692,7 @@ components:
     Url:
       type: object
       properties:
-        path:
+        url:
```
slsriehl, line 21695:
```suggestion
        path:
```
revert for read-only file

slsriehl:
Due to limitations with editing a fork branch and the size of the old catalog file, i'm merging this to a local feature branch so we can revert changes to the read-only file.

bc-traciporter:
Thanks @slsriehl

PR #1311, opened 2023-05-09
add `name` to script_Base

kzhang-dsg:
# [DEVDOCS-]
N/A

## What changed?
* add `name` to script_Base

## Anything else?
N/A


slsriehl:
name should be in `script_Full`, not `script_Base`.

AFAIK name is not required in PUT requests, but is in POST. The limits of OAS led writers to structure this file such that `script_Base` contains shared properties that are not required, and the required fields are added per operation.

PR #1334, opened 2023-05-31
DEVDOCS-4718: [revise] Tax Provider API, wrapping field & allowed types

bc-andreadao:
# [DEVDOCS-4718]

## What changed?
- Fixed incorrect examples (e.g. wrapping line item should only have `type: wrapping` )
- Created a schema for each type of line item so the enum values can be more accurate
  - items: allowed value for its type is `item`
  - shipping: allowed value for its type is `shipping`
  - handling: allowed value for its type is `handling`
  - wrapping: allowed value for its type is `wrapping`
- With some structural changes, only `request-item` and `response-item` are now referenced, so I removed `request-item-wrapping` and `response-item-wrapping`. 

[DEVDOCS-4718]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-4718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/tax_provider.yml
```diff
@@ -1332,6 +1285,42 @@ components:
...
       x-internal: false
+    item_type:
+      type: string
+      description: |-
+        The type of item for the line item in the document.
+
+        Tax estimate requests for order-level refunds use the type `refund`.
+      enum:
+        - item
+        - refund
```
bc-andreadao, line 1296:
```suggestion
        - refund
```
@theromulans  When a request is made for an order-level refund, is `refund` the `type` for all the line items? Is `refund` an allowed value for **each** type of line item (e.g. wrapping line items, handling line items, etc)?

bc-andreadao, line 1296: @theromulans  When a request is made for an order-level refund, is `refund` the `type` for all the line items? Is `refund` an allowed value for **each** type of line item (e.g. wrapping line items, handling line items, etc)?

theromulans: An order-level refund is an additional item outside the bounds of the order - it doesn't affect any other items. It is included as a way of simply informing the tax provider, if they are interested in the information - generally for accounting purposes.

So, yes `refund` is only for order-level refunds, a completely separate item type. Handling, shipping, wrapping cannot be a `refund`. So, less an `enum` there and more an exact value of `handling`, `shipping` or `wrapping`.

reference/tax_provider.yml
```diff
@@ -925,94 +925,10 @@ components:
...
-            - wrapping
-            - handling
-            - shipping
-            - refund
       required:
         - id
         - price
         - quantity
-        - type
-    request-item-wrapping:
```
theromulans, line 945: Loving this more scalable approach, thank you @bc-andreadao !

reference/tax_provider.yml
bc-andreadao: Thanks @theromulans ! I removed `refund` from handling, shipping, and wrapping. Let me know if there's anything else

bc-tgomez (approved):

theromulans:
Thanks @bc-andreadao !

PR #1336, opened 2023-06-01
[WIP]feat(checkout): CHECKOUT-7507 Design Fees Read path API

valentindellangela:
## Jira?
* [CHECKOUT-7507]
* [DEVDOCS-5078]

## What?
Update the API Schema to support the read path of the new Checkout Fees in V3 and Storefront.

This PR is only a Draft that will be used to share the API Design with the Partner `Route`

## Tested?
In [Swagger.io](https://editor.swagger.io/) 
image

Example:
image

ping @bigcommerce/checkout 

[CHECKOUT-7507]: https://bigcommercecloud.atlassian.net/browse/CHECKOUT-7507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEVDOCS-5078]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/checkouts.v3.yml
```diff
@@ -8511,6 +8516,41 @@ components:
...
+          description: Display name of the fee targeting customers/shoppers
+          example: Package Protection Insurance
+        cost_inc_tax:
+          type: number
+          description: Cost of the fee including tax
+          example: 10.0
+        cost_ex_tax:
+          type: number
+          description: Cost of the fee excluding tax
+          example: 9.9
```
bc-shawnwang, line 8548: 👍

bc-shawnwang, line 8548: 👍

animesh1987: Do we need to worry about the inc/ex tax?

bc-shawnwang: I think now we don't, but if we just have a `cost`, later changed to `cost_inc_tax` and `cost_ex_tax`, it will be a breaking change. so better have tax suffix now for future compatibility. @valentindellangela

valentindellangela: https://bigcommerce.slack.com/archives/C057GGG099V/p1685662677660899?thread_ts=1685578267.444999&cid=C057GGG099V

Brett also recommends to handle both `inc_tax` and `ex_tax` as tax will be applied in the background.

animesh1987: I still don't think that is something we should worry with fee providers, they can return a cost and as a result of that we can let our tax domain make the call. Like even if they return a cost ex/inc price it might be different from what a merchant's tax rates are or region tax rates are. For example US has different tax rates per region. 

Therefore for a provider to calculate such prices correctly, they need to know the address and this further complicates things. Hence I feel we should just have a cost that is returned as part of this without complicating things.

animesh1987: Sorry I think there is a bit of confusion, I am no opposed to the idea of inc/ex being returned. What I mainly wanted to point out was that there is no `cost`/`base_cost` field being returned, which I think we might need to consider.

animesh1987: Sep pointed me out to this as a reference
Screenshot 2023-06-02 at 2 25 24 pm

valentindellangela: is it in the Checkout API? 🤔  I can't see it anywhere https://developer.bigcommerce.com/docs/rest-management/checkouts#get-a-checkout

So you're suggesting to have a `base_cost` for now and in the future, we could add `ex_tax` and `inc_tax`?

animesh1987: Yes possibly :D

valentindellangela: What's happening for merchant settings things as tax included and display tax exclusive? 

I think from the discussion with @theromulans my understanding is we apply a tax class in BCAPP. So we can get the cost price exclusive or inclusive of tax. 
It's just that we don't let the client pass the tax class ID.

theromulans: Yes that's right @valentindellangela.

A merchant may change the fee tax class at any time via tax settings and hence all fees can have a tax component.

So, yes, the response needs to handle it - just the same as handling does today.

Does that answer the question?

reference/checkouts.sf.yml
```diff
@@ -2434,6 +2434,41 @@ components:
...
+          description: Display name of the fee targeting customers/shoppers
+          example: Package Protection Insurance
+        cost_inc_tax:
+          type: number
+          description: Cost of the fee including tax
+          example: 10.0
+        cost_ex_tax:
+          type: number
+          description: Cost of the fee excluding tax
+          example: 9.9
```
bc-shawnwang, line 2466: I missed it, seems like the naming case is inconsistent. should be costIncTax and costExTax?

reference/checkouts.v3.yml
```diff
@@ -8511,6 +8516,41 @@ components:
...
+          enum:
+            - custom_fee
+        name:
+          type: string
+          description: Name of the fee.
+          example: Route Package Protection Insurance
+        displayName:
+          type: string
+          description: Display name of the fee targeting customers/shoppers
+          example: Package Protection Insurance
```
bc-shawnwang, line 8540: for s2s should it be display_name?

PR #1361, opened 2023-06-21
DEVDOCS-5141 [update]: Shipping V2 API, Update byweight and bytotal default_cost description

agchan12:
## What changed?

- Updated the `default_cost` property of the static methods `weight` and `total` to support a number or null values

## Why?

We have updated our v2/shipping/zone/:zoneId/methods api for `default_cost` property on our static `weight` and `total` shipping method to support number or null values. This should be reflected in our API documentation.

## Testing/Proof

Proof

ping @bigcommerce/shipping @bc-andreadao


reference/shipping.v2.yml
```diff
@@ -1082,12 +1082,12 @@ paths:
         
         | Property | Type | Description |
         |:---------|:-----|:------------|
-        | default_cost | number | Default shipping cost, applied either as a percentage of the orderʼs total value or as a fixed amount. |
+        | default_cost | number &#124; null | Default shipping cost, applied either as a percentage of the orderʼs total value or as a fixed amount. A value of null can be supplied if no default cost is required. |
```
agchan12, line 1085: For context the `&#124;` is used to display the pipe (`|`) character.

When I tried to escape the `|` character, it still registered the pipe character as another column in the table and the escape char shows

escape_char

bc-andreadao, line 1085:
```suggestion
        | default_cost | number &#124; null | Default shipping cost, applied either as a percentage of the orderʼs total value or as a fixed amount. If default cost is not required, you can supply a value of null. |
```

bc-traciporter (approved):
LGTM!

PR #1370, opened 2023-07-06
DEVDOCS-4881 [update] Orders v2, Update schema for Order Products

bc-tgomez:
# [DEVDOCS-4881]

## What changed?

* Update the missing `quantity_refunded` response for Order Products


bc-traciporter (approved):
LGTM

PR #1422, opened 2023-08-28
Make fixes in cart OpenApi Spec and move some reusable components to one place

bc-tayvs:
# [DEVDOCS-]

## What changed?
Provide a bulleted list in the present tense
* 

## Anything else?
Add related PRs, salient notes, ticket numbers, etc.

ping {names}


reference/carts.v3.yml
```diff
@@ -775,21 +775,23 @@ paths:
       tags:
         - Metafields
       description: Get all cart metafields.
-      operationId: getAllCartMetafields
+      operationId: getAllCartMetafields #getAllMetafieldsByCartId?
```
slsriehl, line 778:
```suggestion
      operationId: getAllCartMetafields
```

slsriehl:
@bc-tayvs -- This is a good idea, but we need to do more work to figure out how moving to external references will affect existing API clients. In addition, we're migrating all the docs that are currently in public repos to https://github.com/bigcommerce/docs. 

So -- keep advocating for external references, I think it's a good idea in the abstract and we need time to discern a path forward on making it a reality while still serving all our devs.

PR #1456, opened 2023-09-06
feat(tax): TAX-1867 Update tax settings documentation for tax zone based price display for storefront

bc-msung:
# [DEVDOCS-]

## What changed?
As part of [Tax Zone Based Price Display for Storefront](https://bigcommercecloud.atlassian.net/browse/PROJECT-3341), three new tax settings are being introduced to the v3 endpoint.
`should_use_geolocation_to_determine_guest_shopper_tax_zone` - enabling this features will allow the merchants to display the tax estimation based on the user's geolocation.
'guest_shopper_tax_zone_id` - default tax zone id for guest shoppers entering the storefront
`store_tax_zone_id` - default tax zone id for store

## Anything else?
Geolocation PR
https://github.com/bigcommerce/bigcommerce/pull/53689
https://github.com/bigcommerce/bigcommerce/pull/53561

Default Tax Zone ID
https://github.com/bigcommerce/bigcommerce/pull/54330
https://github.com/bigcommerce/bigcommerce/pull/54157

ping {names}


reference/tax_settings.v3.yml
```diff
@@ -146,7 +146,19 @@ components:
...
+          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the itemʼs store tax rate before calculating tax using the shopperʼs tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopperʼs location. The tax-exclusive amount may vary among tax zones. These calculations are relevant for tax pricing and tax quotations that use basic tax.
+        should_use_geolocation_to_determine_guest_shopper_tax_zone:
+          default: false
+          type: boolean
+          description: Configures the tax estimation and display settings the shoppers see based on their geolocation before they confirm the destination address at checkout.
+        guest_shopper_tax_zone_id:
+          default: 1
+          type: integer
+          description: Setting to configure the default guest shopper tax zone id.
```
theromulans, line 157: These new descriptions don't help describe the _why_ of these settings at all - care to take another crack @bc-msung?

What is the guest shopper tax zone? What is the store tax zone? Which circumstances are these relevant (i.e. store tax zone ID only applies "if a merchant enters tax-inclusive prices" (matching the should_subtract_store_tax description immediately above)?

reference/tax_settings.v3.yml
```diff
@@ -178,6 +190,18 @@ components:
           default: true
           type: boolean
           description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the itemʼs store tax rate before calculating tax using the shopperʼs tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopperʼs location. The tax-exclusive amount may vary among tax zones. These calculations are relevant for tax pricing and tax quotations that use basic tax.
+        should_use_geolocation_to_determine_guest_shopper_tax_zone:
```
theromulans, line 193: 🍻 Seems a bit funny that we need to describe this twice - is there a better approach?

i.e. `Tax_Settings` and `Tax_Settings_Req` would surely be a match for the majority of the time?

Might be something for @bc-andreadao to advise on.

reference/tax_settings.v3.yml
```diff
@@ -146,7 +146,19 @@ components:
...
+          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the itemʼs store tax rate before calculating tax using the shopperʼs tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopperʼs location. The tax-exclusive amount may vary among tax zones. These calculations are relevant for tax pricing and tax quotations that use basic tax.
+        should_use_geolocation_to_determine_guest_shopper_tax_zone:
+          default: false
+          type: boolean
+          description: Configures the tax estimation and display settings the shoppers see based on their geolocation before they confirm the destination address at checkout.
```
theromulans, line 153: Ah, I thought we were missing something - thanks for including it.

This description could be improved. I note we speak to "When enabled," in the above `should_subtract_store_tax` description, which I think would be a good starting point.

reference/tax_settings.v3.yml
```diff
@@ -146,7 +146,19 @@ components:
...
           type: boolean
-          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the itemʼs store tax rate before calculating tax using the shopperʼs tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopperʼs location. The tax-exclusive amount may vary among tax zones. These calculations are relevant for tax pricing and tax quotations that use basic tax.    
+          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the itemʼs store tax rate before calculating tax using the shopperʼs tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopperʼs location. The tax-exclusive amount may vary among tax zones. These calculations are relevant for tax pricing and tax quotations that use basic tax.
+        should_use_geolocation_to_determine_guest_shopper_tax_zone:
```
agchan12, line 150: I know this has probably been discussed in the PR when introducing this field name, but this name is a mouthful.

Would something shorter like `determine_tax_zone_by_geolocation` be easier to digest here?

reference/tax_settings.v3.yml
bc-msung: updated with more info 🤞

reference/tax_settings.v3.yml
bc-msung: added in `When enabled`

reference/tax_settings.v3.yml
bc-msung: I agree the variable name is quite long, but also think that it is important to keep the context as well. I personally think that keeping `guest_shopper` or related information somewhere in the variable makes it more straight forward for users.
Happy to change it to something else if we get to decide on another variable that is more concise and is clear

theromulans: Yes, we did consider how we could shorten this naming originally but we did not identify a better option.

theromulans:
The PR description looks to need updating, by the way - some of the template items haven't been filled in.

reference/tax_settings.v3.yml
```diff
@@ -146,7 +146,19 @@ components:
...
+          default: false
+          type: boolean
+          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, it allows the store to configure the tax estimation and display settings the shoppers see based on their geolocation before they confirm the destination address at checkout. The tax estimation and calculation will be based on tax zones and rates configured on basic tax.
```
theromulans, line 153: > When enabled, it allows the store to configure

Only when enabled?

reference/tax_settings.v3.yml
```diff
@@ -169,7 +181,7 @@ components:
                 - EXCLUSIVE
         fallback_strategy:
           type: string
-          description: 'Decribes the fallback behaviour that applies when a tax provider produces an error. A merchant may decide to use a flat 10% fallback tax rate, their basic tax settings, or to block the transaction until a successful result can be achieved.'
+          description: 'Describes the fallback behaviour that applies when a tax provider produces an error. A merchant may decide to use a flat 10% fallback tax rate, their basic tax settings, or to block the transaction until a successful result can be achieved.'
```
theromulans, line 184: Nice, thanks for this bonus find 👏

reference/tax_settings.v3.yml
```diff
@@ -178,6 +190,18 @@ components:
...
+        should_use_geolocation_to_determine_guest_shopper_tax_zone:
+          default: false
+          type: boolean
+          description: When enabled, geolocation data will be used to match the guest shopper with a relevant country-level tax zone for tax estimation on the storefront. The geolocation data will be used to determine the corresponding tax zone provided in basic tax - once the corresponding tax zone is found, the tax estimation will be applied to the store. When this setting is disabled, the tax estimation will be based on the guest shopper tax zone, the guest shopper tax zone can be reconfigured by updating guest_shopper_tax_zone_id or store_tax_zone_id.
```
theromulans, line 196: The second sentence looks like a repeat of the first - thoughts?

> the tax estimation will be applied to the store

I'm not sure I'd describe tax estimation as applying to a store.

I think we can simplify this to:

```suggestion
          description: When enabled, geolocation data will be used to match the guest shopper with a relevant country-level tax zone for tax estimation on the storefront. When disabled, storefront tax estimation will always use the guest shopper tax zone.
```

reference/tax_settings.v3.yml
```diff
@@ -178,6 +190,18 @@ components:
...
+        guest_shopper_tax_zone_id:
+          default: 1
+          type: integer
+          description: The setting allows the user to configure the guest shopper tax zone for shoppers to see in the storefront. The user can configure the tax zone id to any existing tax zone in basic tax where the chosen tax zone will be used for tax estimation. When used in conjunction with store_tax_zone_id and should_subtract_store_tax, the rules of should_subtract_store_tax_will apply for the storefront where the tax rates will be subtracted and recalculated based on the two tax zones provided.
```
theromulans, line 200:
```suggestion
          description: Describes which tax zone is used for guest shoppers when requiring storefront tax estimation. When applicable, the shopper's geolocation data will be used in favour of the guest shopper tax zone.
```

It's an interesting challenge to try and provide "just enough" detail, especially since there is a matrix of relationships now. I'm hoping we can limit the store tax subtraction topic to those settings.

reference/tax_settings.v3.yml
```diff
@@ -178,6 +190,18 @@ components:
...
+        store_tax_zone_id:
+          default: 1
+          type: integer
+          description: This setting applies only if a merchant enters tax-inclusive prices and has should_subtract_store_tax enabled. This setting lets the user to configure their own store tax zone in basic tax. When the shopper enters a destination address or provides their own geolocation, the rules of should_subtract_store_tax will apply with regards to tax estimation. This correlates to the store tax rate description provided in the should_subtract_store_tax where the initial tax rates are calculated in the tax zone id provided in this configuration.
```
theromulans, line 204:
```suggestion
          description: This setting applies only if a merchant enters tax-inclusive prices. Describes which tax zone to use for store tax subtraction.
```

I think we can avoid some detail here/leave it for other sections 🙏

reference/tax_settings.v3.yml
theromulans: Thank you sir!

reference/tax_settings.v3.yml
bc-msung: updated 👍

reference/tax_settings.v3.yml
```diff
@@ -146,7 +146,19 @@ components:
...
+          description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the itemʼs store tax rate before calculating tax using the shopperʼs tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopperʼs location. The tax-exclusive amount may vary among tax zones. These calculations are relevant for tax pricing and tax quotations that use basic tax.
+        should_use_geolocation_to_determine_guest_shopper_tax_zone:
+          default: false
+          type: boolean
+          description: When enabled, geolocation data will be used to match the guest shopper with a relevant country-level tax zone for tax estimation on the storefront. The geolocation data will be used to determine the corresponding tax zone provided in basic tax - once the corresponding tax zone is found, the tax estimation will be applied to the store.
```
benpratt77, line 153: Do we need the word data in here?

reference/tax_settings.v3.yml
bc-msung: 👍

reference/tax_settings.v3.yml
bc-msung: removed 👍

theromulans: > geolocation will be used

Doesn't sound quite right to me though 🤔 Do we need a more complete rewording? e.g. "the guest shopper's geolocation will be used to determine the relevant country-level tax zone for storefront tax estimation".

I also wonder if `geolocation` may suggest a complete geolocation, something with a higher accuracy than we are offering, whereas `geolocation data` is more vague and reflects that we're only using a subset of the data.

What do you think @benpratt77?

reference/tax_settings.v3.yml
```diff
@@ -178,6 +190,18 @@ components:
...
+          type: boolean
+          description: When enabled, geolocation will be used to match the guest shopper with a relevant country-level tax zone for tax estimation on the storefront. The geolocation data will be used to determine the corresponding tax zone provided in basic tax - once the corresponding tax zone is found, the tax estimation will be applied to the store.
+        guest_shopper_tax_zone_id:
+          default: 1
+          type: integer
+          description: Describes which tax zone is used for guest shoppers when requiring storefront tax estimation. When applicable, the shopper's geolocation data will be used in favour of the guest shopper tax zone.
+        store_tax_zone_id:
+          default: 1
+          type: integer
+          description: This setting applies only if a merchant enters tax-inclusive prices. Describes which tax zone to use for store tax subtraction.
```
theromulans, line 204:
```suggestion
          description: This setting applies only if a merchant enters tax-inclusive prices and store tax subtraction is enabled. Describes which tax zone to use for store tax subtraction.
```

Thanks for the suggestion here @bc-msung, great call. Good to, in context, identify when this setting is relevant, guiding the user towards using/not using it as relevant.

reference/tax_settings.v3.yml
theromulans: Exploring the language we're using in these descriptions, would this make sense: `should_use_guest_shopper_geolocation`? I'm not completely sold on it, but it does tighten things up.

The flaw is that we're not describing what this is being used for 😆 There could be a fair argument that this is a detail that can be described instead.

reference/tax_settings.v3.yml
benpratt77: Perhaps a more complete rewording. data is so vague. 
So what are we actually using, we are using their IP to determine their country. With the idea that some time in the future we may also use subdivision. (I don't think we will ever try for postal code as that is entirely too inaccurate from my previous digging)

reference/tax_settings.v3.yml
```diff
@@ -178,6 +190,18 @@ components:
...
+        should_use_geolocation_to_determine_guest_shopper_tax_zone:
+          default: false
+          type: boolean
+          description: When enabled, geolocation will be used to match the guest shopper with a relevant country-level tax zone for tax estimation on the storefront. The geolocation data will be used to determine the corresponding tax zone provided in basic tax - once the corresponding tax zone is found, the tax estimation will be applied to the store.
+        guest_shopper_tax_zone_id:
+          default: 1
+          type: integer
+          description: Describes which tax zone is used for guest shoppers when requiring storefront tax estimation. When applicable, the shopper's geolocation data will be used in favour of the guest shopper tax zone.
```
theromulans, line 200:
```suggestion
          description: Describes which tax zone is used for guest shoppers when requiring storefront tax estimation. When applicable, the guest shopper's geolocation will be used in favour of the guest shopper tax zone.
```

- Consistent language for "guest shopper's geolocation", as per above `geolocation data` discussion.
- Removed `data` as above also.

reference/tax_settings.v3.yml
theromulans: @benpratt77 Are you satisfied with my rewording suggestion, or do you have another one to offer?

reference/tax_settings.v3.yml
benpratt77: Does it need to be tight?. I feel like `should_use_geolocation_to_determine_guest_shopper_tax_zone` is very descriptive and is non-ambiguous.

reference/tax_settings.v3.yml
benpratt77: I am satisfied with your rewording.

reference/tax_settings.v3.yml
theromulans: It doesn't need to be.

I was also exploring trying to "proper noun" this feature a bit more, but I'm coming up blank.

reference/tax_settings.v3.yml
bc-msung: 👍

reference/tax_settings.v3.yml
bc-msung: updated 👍

theromulans (approved):
Thank you sir

reference/tax_settings.v3.yml
bc-andreadao: Good point, I removed `Tax_Settings_Req` 👍

reference/tax_settings.v3.yml
```diff
@@ -143,41 +143,22 @@ components:
...
-            - BASIC
-            - DISABLE
         should_subtract_store_tax:
           default: true
           type: boolean
           description: This setting applies only if a merchant enters tax-inclusive prices. When enabled, the store subtracts the itemʼs store tax rate before calculating tax using the shopperʼs tax zone. The tax-exclusive amount will be the same across all tax zones. When disabled, the tax-inclusive price remains the same across all tax zones; only the tax amount will vary based on the shopperʼs location. The tax-exclusive amount may vary among tax zones. These calculations are relevant for tax pricing and tax quotations that use basic tax.
+        should_use_geolocation_to_determine_guest_shopper_tax_zone:
+          default: false
+          type: boolean
+          description: When enabled, the guest shopper's geolocation will be used to determine the relevant country-level tax zone for storefront tax estimation. The geolocation data will be used to determine the corresponding tax zone provided in basic tax - once the corresponding tax zone is found, the tax estimation will be applied to the store.
```
bc-andreadao, line 153:
```suggestion
          description: This setting sets whether a store uses a guest shopper's geolocation to estimate storefront tax. When enabled, the store identifies a country-level tax zone based on the geolocation. The store then applies the corresponding guest shopper's tax zone to tax estimations. When disabled, the store uses the `guest_shopper_tax_zone_id` field to determine the corresponding zone. 
```

reference/tax_settings.v3.yml
```diff
@@ -143,41 +143,22 @@ components:
...
+        guest_shopper_tax_zone_id:
+          default: 1
+          type: integer
+          description: Describes which tax zone is used for guest shoppers when requiring storefront tax estimation. When applicable, the guest shopper's geolocation will be used in favour of the guest shopper tax zone.
```
bc-andreadao, line 157:
```suggestion
          description: Describes which tax zone a store uses to estimate storefront tax for guest shoppers. When applicable, a store uses the guest's geolocation in favor of the guest shopper's tax zone.
```

reference/tax_settings.v3.yml
```diff
@@ -143,41 +143,22 @@ components:
...
+        store_tax_zone_id:
+          default: 1
+          type: integer
+          description: This setting applies only if a merchant enters tax-inclusive prices and store tax subtraction is enabled. Describes which tax zone to use for store tax subtraction.
```
bc-andreadao, line 161:
```suggestion
          description:  This setting applies only if a merchant 1) enters tax-inclusive prices and 2) subtracts store tax before tax calculation. Describes which tax zone to use when subtracting store tax.
```

bc-andreadao:
@bc-msung I made some suggestions. I believe they are still accurate, but let me know.

(I also removed `Tax_Settings_Req` because it pretty much duplicates `Tax_Settings`).

reference/tax_settings.v3.yml
theromulans: > to estimate storefront tax

Sounds a bit awkward to me in that there's no such thing as "storefront tax". Perhaps this speaks to confused language around `store` vs `storefront` too. Perhaps we can simplify slightly:

> This setting determines which tax zone a store uses to estimate tax for guest shoppers.

theromulans: > This setting applies to basic tax; tax zones must be provided in the basic tax set up.

This could be confused as not being relevant for automatic tax providers somehow. What do you think of dropping this sentence and leaving it as implied?

Alternatively a slight simplification to remove the `applies to basic tax`:

> Note: tax zones must be configured in the basic tax set up.

Replacing `provided` with `configured` at least seems like a minor improvement (given we have 'tax providers' as nearby terminology)🤔

reference/tax_settings.v3.yml
theromulans: Similar to the above, we'd probably want to drop `storefront tax`.

> ID for the tax zone a store uses when estimating tax for guest shoppers.

reference/tax_settings.v3.yml
theromulans: Is the inline numbering a thing we regularly do? Looks a bit odd to me.

Would this read better if we lead with:

> ID for the tax zone a store uses when subtracting store tax.

I'm not sure about `field` vs `setting` - I think of `field` as very much at the API layer only. There's no restriction on the user adjusting the field itself, but rather the setting.

bc-andreadao: Good points, just added them in 👍

reference/tax_settings.v3.yml
bc-andreadao: Sounds good, adjusted the wording. I just added in a point about the setting being disabled. I assume there might be times when a guest's geolocation is unavailable even with the setting enabled, unless I'm mistaken

reference/tax_settings.v3.yml
bc-andreadao: Good points, I just removed `storefront` and used `configured` instead.

I see. At first, I didn't realize that Tax Providers could also use geolocation. I'm thinking of:

> If the store uses basic tax, only the tax zones you configure are available.

The sentence emphasizes that it doesn't have to be basic tax. I also reworded it to "only the tax zones you configure are available" to make the meaning slightly more accurate

theromulans: Thanks @bc-andreadao.

It's awkward since all stores use basic tax, for tax pricing.

I understand that the main reason for including this additional note is that we aren't magically pulling numbers out of thin air, we want to guide the merchant that they need to define these (country-level) tax zones that we'll match the guest shopper to.

What about:

> Only the tax zones you configure can be matched to the guest shopper's geolocation.

🤔

reference/tax_settings.v3.yml
theromulans: I'm uncertain about the scenario where a geolocation is unavailable - as far as we've discussed previously we'll always have a value here and hence a match.

The scenarios where this tax zone will be used are:
1) When the geolocation feature is disabled.
2) When the geolocation feature is enabled AND when there is no matching country-level tax zone for the geolocation (e.g. A merchant hasn't set up a Germany tax zone, but we've identified a Germany-based guest shopper via geolocation).

reference/tax_settings.v3.yml
bc-andreadao: Thanks, sounds good 👍

reference/tax_settings.v3.yml
bc-andreadao: Got it, thanks, I revised the sentence 👍

reference/tax_settings.v3.yml
```diff
@@ -143,41 +143,22 @@ components:
...
+        should_use_geolocation_to_determine_guest_shopper_tax_zone:
+          default: false
+          type: boolean
+          description: This setting determines which tax zone a store uses to estimate tax for guest shoppers. When enabled, the store identifies a country-level tax zone based on the geolocation of a guest shopper. The store then applies the corresponding tax zone to estimate taxes. When disabled, the store identifies the zone using the provided `guest_shopper_tax_zone_id` field instead. Only the tax zones you configure can be matched to the guest shopper's geolocation.
+        guest_shopper_tax_zone_id:
+          default: 1
+          type: integer
+          description: ID for the tax zone a store uses when estimating tax for guest shoppers. The store uses this zone if you disable `should_use_geolocation_to_determine_guest_shopper_tax_zone`. The store also uses this zone if there is no matching country-level tax zone for the geolocation.
```
bc-tgomez, line 157:
```suggestion
          description: ID for the tax zone that the store is set to use when estimating tax for guest shoppers. The geo-location that the store relies on when `should_use_geolocation_to_determine_guest_shopper_tax_zone` is enabled. The default for any store relies on this zone if there is no matching country-level tax zone for the geolocation.
```
You don't have to use this verbatim, just trying to rearrange some of the words to make it less like the store performs things like a person.

bc-tgomez (commented):
I made a comment but other than that this is well done!

reference/tax_settings.v3.yml
bc-andreadao: I might leave it because the meaning changes and it's less concise

bc-traciporter (approved):
Good job @bc-andreadao

PR #1465, opened 2023-09-11
DEVDOCS-5408 [new]: Catalog API, add sort=name query parameter

bc-andreadao:
# [DEVDOCS-5408]

## What changed?
- Add sort=name as query parameter to Get All Categories endpoint

cc @kristinapototska


reference/catalog/categories_catalog.v3.yml
```diff
@@ -215,6 +215,11 @@ paths:
           description: 'Fields to exclude, in a comma-separated list. The specified fields will be excluded from a response. The ID cannot be excluded.'
           schema:
             type: string
+        - name: sort
+          in: query
+          description: 'Controls the sort order of the response. For example, `sort=name` sorts categories in alphabetical order by `name`.'
```
jyuricle, line 220: I think we need to specify a list of all allowed values to sort by. In addition to `name` we have `id`, `sort_order` and `parent_id`

bc-andreadao: Thanks @jyuricle , I added the allowed values

bc-traciporter (approved):
LGTM!

PR #1469, opened 2023-09-17
Update Operation IDs to be consistent & programmatic

bobbyshaw:
# [DEVDOCS-]

As per OpenAPI specification guidance for operationID field, follow programmatic
conventions for operationId field. To me that means as if I were defining a method
or function name.

Screenshot 2023-09-17 at 20 44 14

This field is not used in the API docs as far as I can tell. There were
many different naming conventions across APIs. I've aligned to be camel case.

This help facilitate API SDK generation as all endpoints can have a function named
based on the operation ID and it have sufficient context to be unique amongst all APIs

## What changed?
Provide a bulleted list in the present tense
* Make operation IDs consistent and follow programmatic conventions

### DIFF
--- a/reference/abandoned_cart_emails.v3.yaml
+++ b/reference/abandoned_cart_emails.v3.yaml
@@ -57,11 +57,11 @@ paths:
                     meta: {}
       tags:
         - Abandoned Cart Emails
-      summary: Get all email templates
-      operationId: getAbandonedCartEmails
+      summary: Get all abandoned cart email templates
+      operationId: getAbandonedCartEmailTemplates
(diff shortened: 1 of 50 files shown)


slsriehl:
@bobbyshaw this is tremendous. thank you!

@markcmurphy, @bc-shawnyap, @becomevocal -- can you confirm that this will not break the parser, the request runner, or any other Dev Center functionality?

slsriehl:
See successor branch https://github.com/bigcommerce/docs/tree/bobbyshaw-consistent-operation-ids.

PR #1490, opened 2023-10-06
DEVDOCS-5531: [revise]Update consignment endpoints

bc-traciporter:
# [DEVDOCS-5531]

## What changed?
* Removed the array in the Update Checkout consignment and changed it to an object.
* Changed `address` to `shipping address` in the Add Consignment to Checkout
* Updated notes in the Update Checkout consignment endpoint about updating shipping options and addresses separately.

## Anything else?
Add related PRs, salient notes, ticket numbers, etc.

ping {names}

[DEVDOCS-5531]: https://bigcommercecloud.atlassian.net/browse/DEVDOCS-5531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/checkouts.v3.yml
```diff
@@ -8619,7 +8622,7 @@ components:
       items:
         type: object
         properties:
-          address:
+          shipping_address:
```
bc-shawnwang, line 8625: I remember we allow both `address` and `shipping_address` (we defined them as alias). am I right? @valentindellangela

reference/checkouts.v3.yml
```diff
@@ -8703,10 +8706,8 @@ components:
       x-internal: false
     UpdateConsignmentRequest:
       title: Update Consignment Request
-      type: array
-      items:
-        type: object
-        properties:
+      type: object
```
bc-shawnwang, line 8709: thanks for the change, this looks correct for me

reference/checkouts.v3.yml
bc-shawnwang: image

`shipping_address` (deprecated) or `address`

This is from storefront doc description. I checked the s2s part, it should be the same. and we encourage to use address

bc-tgomez (approved):
Looks good!

PR #1527, opened 2023-11-21
(no ticket): [revise] query param include should accept comma separated list of values

matthewvolk:
## What changed?

_For context, this PR originated from: https://github.com/matthewvolk/bigrequest/issues/84 as a solution for inconsistent Typescript type definitions being generated from our OpenAPI spec files. I haven't checked if these changes introduce any unintended side effects to other ways our spec files are consumed!_

- Refactors the schema definition of the `include` query parameter across various API endpoints to indicate that `include` accepts a comma-separated list of enum values, rather than one-of an enum value. 
- A schema with `type: string` and `enum: [value1,value2]` resolves to: `?include=[value1|value2]` (_either_ `value1` _or_ `value2`) ([source: swagger.io](https://swagger.io/docs/specification/data-models/enums/))
- If we instead describe the schema as a `type: array`, where the `items` in the array are of `type: string` and can be one or more of `items.enum: [value1,value2]` the schema now resolves to: `?include=[value1[,value2]`
- `explode: false` is also added to indicate that query parameter values can be expressed in the format: `/users?id=3,4,5` as opposed to `/users?id=3&id=4&id=5` ([source: swagger.io)](https://swagger.io/docs/specification/serialization/#query))

## Anything else?

There are a number of locations in other files that we may want to apply these changes to, but I haven't confirmed that each location accepts a comma separated list of strings. Here is a list: https://github.com/search?q=repo%3Abigcommerce%2Fapi-specs%20%22-%20name%3A%20include%22&type=code

### DIFF
--- a/reference/customers.v3.yml
+++ b/reference/customers.v3.yml
@@ -171,13 +171,18 @@ paths:
              * `segment_ids`- segments the customer belongs to (Beta)
 
              `include=addresses,storecredit,attributes,formfields,shopper_profile_id,segment_ids`
+          explode: false
           schema:
-            type: string
-            enum:
-              - addresses
-              - storecredit
-              - attributes
-              - formfields
+            type: array
+            items:
+              type: string
+              enum:
+                - addresses
+                - storecredit
+                - attributes
+                - formfields
+                - shopper_profile_id
+                - segment_ids
         - in: query
           name: sort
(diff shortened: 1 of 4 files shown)


reference/customers.v3.yml
```diff
@@ -171,13 +171,18 @@ paths:
...
+            type: array
+            items:
+              type: string
+              enum:
+                - addresses
+                - storecredit
+                - attributes
+                - formfields
+                - shopper_profile_id
+                - segment_ids
```
matthewvolk, line 185: These two values were mentioned in the description, but were not included in the enum. I can remove if the description wasn't meant to include them.

matthewvolk:
fixes matthewvolk/bigrequest#84

reference/customers.v3.yml
slsriehl: let's leave them.

slsriehl (approved):
thank @matthewvolk !

PR #1537, opened 2023-12-11
Updated coupon name description

gslandtreter:
Creating a second coupon with the same name as an existing coupon will cause the API to return `409 Conflict`. Hence, the name parameter must be unique.


bc-tgomez:
Thanks for the feedback, @gslandtreter! We'll merge this after getting approval from @bc-traciporter. 🙇

reference/marketing.v2.yml
```diff
@@ -943,7 +943,7 @@ components:
       properties:
         name:
           type: string
-          description: The name of the coupon.
+          description: The name of the coupon. Value must be unique.
```
bc-traciporter, line 946:
```suggestion
          description: The name of the coupon. The value must be unique.
```

bc-traciporter (dismissed):
I left one comment. You can take it or leave it.

slsriehl (approved):
re-approving after accepting @bc-traciporter 's suggestion.

PR #1546, opened 2024-01-09
feat(checkout): CHECKOUT-7899 Update docs of s2s cart api to allow passing source

bc-shawnwang:
- [ ] wait the api change reaches all tiers

# [DEVDOCS-]

## What changed?
s2s api (v3) POST /carts now can accept an optional field `source` in the payload. only one value now are accepted.

Screenshot 2024-01-09 at 1 22 59 pm

## Anything else?

ping @bigcommerce/team-checkout


reference/carts.v3.yml
```diff
@@ -1134,6 +1134,11 @@ components:
           description: 'The locale of the cart. Accepts strings of format `xx` or `xx-YY`. Uses the [ISO-639 standard](https://www.iso.org/iso-639-language-codes.html) format.'
           format: ISO-639
           example: en-US
+        source:
```
valentindellangela, line 1137: Do we want to expose this publicly / allow developer to call V3 to create Manual Order directly?
I think this will be used by CP only so maybe we don't need to document it for now.

Also I would say it can be confusing to have this in the doc as we don't want people to believe they need to send the `source` on every requests, especially if we say we only accept `MANUAL_INCOMPLETE`. 

What do you think?

bc-shawnwang: yes that makes sense. I think again, that keep this a secret now should not cause any issue.

